Issue #22, opened 2015-04-15
Shorten names by removing ApplicationInsights from class names


olegsych:
Including ApplicationInsights in class names is redundant because the product name is already included in the namespace. Making the following naming changes will help to both shorten the component names and make them more descriptive.

ApplicationInsightsExceptionMiddleware -> ExceptionTrackingMiddleware
ApplicationInsightsRequestMiddleware -> RequestTrackingMiddleware

PR #45, opened 2015-04-24
Make names of middleware classes shorter and more descriptive

olegsych:
To address issue #22, this change removes term "ApplicationInsights" from the names of our middleware classes because it already appears in the namespace. It also adds term "Tracking" to make middleware class names more descriptive.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNet/ExceptionTrackingMiddleware.cs
+++ b/src/Microsoft.ApplicationInsights.AspNet/ExceptionTrackingMiddleware.cs
@@ -5,12 +5,12 @@
     using System;
     using System.Threading.Tasks;
 
-    public sealed class ApplicationInsightsExceptionMiddleware
+    public sealed class ExceptionTrackingMiddleware
     {
         private readonly RequestDelegate next;
         private readonly TelemetryClient telemetryClient;
 
-        public ApplicationInsightsExceptionMiddleware(RequestDelegate next, TelemetryClient client)
+        public ExceptionTrackingMiddleware(RequestDelegate next, TelemetryClient client)
         {
             this.next = next;
             this.telemetryClient = client;
--- a/src/Microsoft.ApplicationInsights.AspNet/ExceptionTrackingMiddleware.cs
+++ b/src/Microsoft.ApplicationInsights.AspNet/ExceptionTrackingMiddleware.cs
@@ -1,10 +1,13 @@
 ﻿namespace Microsoft.ApplicationInsights.AspNet
 {
-    using Microsoft.AspNet.Builder;
-    using Microsoft.AspNet.Http;
     using System;
     using System.Threading.Tasks;
+    using Microsoft.AspNet.Builder;
+    using Microsoft.AspNet.Http;
 
+    /// <summary>
+    /// Sends telemetry about exceptions thrown by the application to the Microsoft Application Insights service.
+    /// </summary>
     public sealed class ExceptionTrackingMiddleware
     {
         private readonly RequestDelegate next;
--- a/src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
@@ -26,7 +26,7 @@
 
         public static IApplicationBuilder UseApplicationInsightsExceptionTelemetry(this IApplicationBuilder app)
         {
-            app.UseMiddleware<ApplicationInsightsExceptionMiddleware>();
+            app.UseMiddleware<ExceptionTrackingMiddleware>();
             return app;
         }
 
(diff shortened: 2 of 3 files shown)


src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -17,17 +17,12 @@ public static class ApplicationInsightsExtensions
     {
         public static IApplicationBuilder UseApplicationInsightsRequestTelemetry(this IApplicationBuilder app)
```
olegsych: What do you think about renaming this method to TrackRequestTelemetry?

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -17,17 +17,12 @@ public static class ApplicationInsightsExtensions
...
         {
-            // TODO: Register if customer did not register
-            //app.UseRequestServices();
-
-            app.UseMiddleware<ApplicationInsightsRequestMiddleware>();
-            return app;
+            return app.UseMiddleware<RequestTrackingMiddleware>();
         }
 
         public static IApplicationBuilder UseApplicationInsightsExceptionTelemetry(this IApplicationBuilder app)
```
olegsych: What do you think about renaming this method to TrackExceptionTelemetry?

src/Microsoft.ApplicationInsights.AspNet/ExceptionTrackingMiddleware.cs
```diff
@@ -5,12 +5,12 @@
     using System;
     using System.Threading.Tasks;
 
-    public sealed class ApplicationInsightsExceptionMiddleware
+    public sealed class ExceptionTrackingMiddleware
```
olegsych: What do you think about term "Telemetry" instead of "Tracking" in the names of exception and request middleware?

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
SergKanz: I believe "Use" is a standard prefix indicating that middleware will be injected

SergKanz:
Looks good.

upendras:
:shipit:

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
Eilon: Right, the `Use` word should be used (har-har!) for any extension method that adds middleware to the builder pipeline.

olegsych: Thanks, @Eilon. Since the `Use` prefix is important, the raw `app.UseMiddleware<RequestTrackingMiddleware>()` call feels more clear and descriptive than the `app.UseApplicationInsightsRequestTelemetry()` we have now. Why do we need to provide an Application Insights-specific extension method?

Eilon: For Intellisense discoverability.

olegsych: Is that why extension methods are placed in the `Microsoft.Framework.DependencyInjection` namespace? Do we need the term `ApplicationInsights` in the method name because the extension method is defined outside of the `Microsoft.ApplicationInsights` namespace?
Would `UseRequestTrackingMiddleware` be a better method name than `UseApplicationInsightsRequestTelemetry` or is the term `Middleware` implied here by the presence of the `Use` prefix?

SergKanz: For the second question - we discussed whether to use ApplicationInsights term in the name or not even when it was in our own namespace. The main idea of using it was to increase code readability - you need to understand that this is Application Insights telemetry that this middleware is collecting. Same way as Facebook authentication has Facebook in the name even if it will be placed in Facebook namespace.

Eilon: It's certainly debatable. It's indeed a matter of how likely you think someone is to confuse this with something else or otherwise not understand what it is. The term "Middleware" in the method name is not required - it's implied.

olegsych: OK. The question is closed. Thanks for sharing your thoughts on this, @Eilon & @SergKanz.

Issue #35, opened 2015-04-20
Unhandled exceptions are collected with handledAt = UserCode instead of Platform


abaranch:
"data": {
    "baseType": "ExceptionData",
    "baseData": {
      "ver": 2,
      "handledAt": "UserCode",
      "properties": {
        "DeveloperMode": "true"
      },

PR #46, opened 2015-04-24
Change exception tracking to report it as HandledAt = Platform

olegsych:
Related issue #35.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsExceptionMiddleware.cs
+++ b/src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsExceptionMiddleware.cs
@@ -1,9 +1,10 @@
 ﻿namespace Microsoft.ApplicationInsights.AspNet
 {
-    using Microsoft.AspNet.Builder;
-    using Microsoft.AspNet.Http;
     using System;
     using System.Threading.Tasks;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.AspNet.Builder;
+    using Microsoft.AspNet.Http;
 
     public sealed class ApplicationInsightsExceptionMiddleware
     {
@@ -26,7 +27,9 @@
             {
                 if (this.telemetryClient != null)
                 {
-                    this.telemetryClient.TrackException(exp);
+                    var exceptionTelemetry = new ExceptionTelemetry(exp);
+                    exceptionTelemetry.HandledAt = ExceptionHandledAt.Platform;
+                    this.telemetryClient.Track(exceptionTelemetry);
                 }
 
                 throw;
--- a/src/Microsoft.ApplicationInsights.AspNet/Fakes.cs
+++ b/src/Microsoft.ApplicationInsights.AspNet/Fakes.cs
@@ -1,6 +1,7 @@
 ﻿#if fakeportable
 using Microsoft.ApplicationInsights.Channel;
 using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
 using System;
 using System.Collections.Generic;
 
@@ -8,6 +9,14 @@ namespace Microsoft.ApplicationInsights
 {
     public class TelemetryClient
     {
+        public TelemetryClient()
+        {
+        }
+
+        public TelemetryClient(TelemetryConfiguration telemetryConfiguration)
+        {
+        }
+
         public TelemetryContext Context { get; }
         public void TrackEvent(string eventName)
         {
@@ -37,6 +46,9 @@ namespace Microsoft.ApplicationInsights
         {
         }
         public void TrackTrace(string message, SeverityLevel? severityLevel, IDictionary<string, string> properties)
+        {
+        }
+        public void Track(ITelemetry telemetry)
         {
         }
     }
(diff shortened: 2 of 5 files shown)


src/Microsoft.ApplicationInsights.AspNet/Fakes.cs
```diff
@@ -1,6 +1,7 @@
 ﻿#if fakeportable
```
olegsych: Notice the amount of changes is this file was large compared to the changes in product and test code. I hope we can drop this soon.

src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsExceptionMiddleware.cs
```diff
@@ -26,7 +27,9 @@ public ApplicationInsightsExceptionMiddleware(RequestDelegate next, TelemetryCli
             {
                 if (this.telemetryClient != null)
```
olegsych: This condition will never be false. If the TelemetryClient service is not available, construction of this object will fail and the Invoke method will not be called. I'd like to remove the if statement. Any objections?

test/Microsoft.ApplicationInsights.AspNet.Tests/FakeTelemetryChannel.cs
```diff
@@ -6,6 +6,8 @@
     // TODO: Remove FakeTelemetryChannel when we can use a dynamic test isolation framework, like NSubstitute or Moq
```
olegsych: Hopefully we can start using a dynamic test isolation framework like Moq or NSubstitute and stop maintaining the fakes in tests.

src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsExceptionMiddleware.cs
```diff
@@ -26,7 +27,9 @@ public ApplicationInsightsExceptionMiddleware(RequestDelegate next, TelemetryCli
...
                 {
-                    this.telemetryClient.TrackException(exp);
+                    var exceptionTelemetry = new ExceptionTelemetry(exp);
+                    exceptionTelemetry.HandledAt = ExceptionHandledAt.Platform;
+                    this.telemetryClient.Track(exceptionTelemetry);
                 }
 
                 throw;
```
olegsych: Technically, the exception is unhandled and rethrown. Shouldn't we report it as HandledAt=Unhandled? What if there is no other middleware that handles exceptions?

test/Microsoft.ApplicationInsights.AspNet.Tests/ApplicationInsightsExceptionMiddlewareTest.cs
```diff
@@ -0,0 +1,38 @@
...
+            RequestDelegate nextMiddleware = httpContext => { throw new Exception(); };
+            var middleware = new ApplicationInsightsExceptionMiddleware(nextMiddleware, MockTelemetryClient());
+
+            await Assert.ThrowsAnyAsync<Exception>(() => middleware.Invoke(null));
+
+            Assert.Equal(ExceptionHandledAt.Platform, ((ExceptionTelemetry)sentTelemetry).HandledAt);
+        }
+
+        private TelemetryClient MockTelemetryClient()
+        {
```
olegsych: This is a lot of code to mock a TelemetryChannel. It would be nice to have an interface with a single Track(ITelemetry) method.

olegsych:
@abaranch, @SergKanz, I have a few questions. Could you take a look at the comments? :cherries:

src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsExceptionMiddleware.cs
SergKanz: sure

src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsExceptionMiddleware.cs
SergKanz: yes

olegsych: What do you mean by "yes"? :smile:

test/Microsoft.ApplicationInsights.AspNet.Tests/FakeTelemetryChannel.cs
SergKanz: Maybe Moq is working with VS RC. If you will not get to it, I'll try it when will have time

SergKanz:
Looks good. Approved

src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsExceptionMiddleware.cs
abaranch: Having Unhadled will make it inconsistent to what is done in web sdk.

test/Microsoft.ApplicationInsights.AspNet.Tests/FakeTelemetryChannel.cs
olegsych: I just tried NSubstitute, which works with dnx451. I'm pretty sure it's because it supports framework net45.Moq will be the same. We can start using either one if we remove the dnxcore50 framework from our test projects. Do we want to do that?

abaranch: Does that mean having no tests for Core CLR at all?

SergKanz: I believe Moq was compiling on both for me. I have limited experience with those. I'd prefer to test on both targets unless it significantly worse.

olegsych: @abaranch, as far as I can tell our tests are not running on dnxcore50 even now.
@SergKanz, Moq has the same problem - it supports net40, but not dnxcore50.

src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsExceptionMiddleware.cs
olegsych: Done.

src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsExceptionMiddleware.cs
olegsych: OK. Leaving it as "Platform".

PR #49, opened 2015-04-27
Sergkanz/remove scripts

SergKanz:
These javascript files will be downloaded automatically


.gitignore
```diff
@@ -22,7 +22,9 @@ nuget.exe
...
 *.*sdf
 *.ipch
-bower_components/
-node_modules/
 .vs/
 project.lock.json
+
+bower_components/
+node_modules/
+**/wwwroot/lib
```
olegsych: Is this the gist of the change in this pullrequest (removing wwwroot/lib from source control)?

SergKanz: I'm removing wwwroot/lib so it will not show up every time npm will download scripts. Ordering change is just for grouping

abaranch:
Approved

PR #50, opened 2015-04-28
Add unit tests for request name and some tests refactoring

abaranch:


SergKanz:
Approved

test/Microsoft.ApplicationInsights.AspNet.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
```diff
@@ -72,16 +71,107 @@ public void InitializeSetsTelemetryOperationNameToMethodAndPath()
...
             initializer.Initialize(telemetry);
 
             Assert.Equal("GET /Test", telemetry.Name);
         }
+
+        [Fact]
+        public void InitializeSetsTelemetryOperationNameToControllerFromActionContext()
+        {
+            var actionContext = new ActionContext();
+            actionContext.RouteData = new Microsoft.AspNet.Routing.RouteData();
```
olegsych: Why do we use the full namespace here instead of a using directive at the top of the file?

olegsych:
Shouldn't this merge to the develop branch instead of master?

test/Microsoft.ApplicationInsights.AspNet.Tests/Helpers/HttpContextAccessorHelper.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.AspNet.Http.Core;
+using Microsoft.AspNet.Mvc;
+using Microsoft.Framework.DependencyInjection;
+using System;
+
+namespace Microsoft.ApplicationInsights.AspNet.Tests.Helpers
+{
+    public static class HttpContextAccessorHelper
+    {
+        public static HttpContextAccessor CreateHttpContextAccessor(RequestTelemetry requestTelemetry = null, ActionContext actionContext = null)
```
olegsych: I like how this new method makes tests shorter. On the other hand, it's also a good example of why *Service Locator* is an anti-pattern. 
- This logic is quite complex for something as simple as a telemetry initializer test.
- The `RequestTelemetry` and `ActionContext` dependencies are hidden in the `RequestServices` service provider.
- The tests would be much simpler and clearer if the components under tests used constructor injection instead.

test/Microsoft.ApplicationInsights.AspNet.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
```diff
@@ -72,16 +71,107 @@ public void InitializeSetsTelemetryOperationNameToMethodAndPath()
...
+            actionContext.RouteData.Values.Add("parameter", "myName");
+
+            var contextAccessor = HttpContextAccessorHelper.CreateHttpContextAccessor(new RequestTelemetry(), actionContext);
+
+            var initializer = new OperationNameTelemetryInitializer(contextAccessor);
+
+            var telemetry = new EventTelemetry();
+            initializer.Initialize(telemetry);
+
+            Assert.Equal("GET account/login [parameter]", telemetry.Context.Operation.Name);
```
olegsych: This test *appears* to duplicate the previous two. I do think it is necessary because parameter logic is predicated on presence of action and controller. These three tests would be easier to understand if each test asserted only on one thing - first on controller, second on action and third on parameter. For instance, this test could be changed to verify only parameter name by 
- Using default(string) or "any" instead of specifying controller and action names.
- Replacing Assert.Equal with Assert.EndsWith("[parameter]", telemetry.Context.Operation.Name)

test/Microsoft.ApplicationInsights.AspNet.Tests/Helpers/HttpContextAccessorHelper.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Core;
+using Microsoft.AspNet.Mvc;
+using Microsoft.Framework.DependencyInjection;
+using System;
```
olegsych: Since we are staying with the Core coding style, using directives need to be sorted system-first and placed inside of the namespace.

test/Microsoft.ApplicationInsights.AspNet.Tests/Helpers/HttpContextAccessorHelper.cs
abaranch: It Is fine that it is hidden. This method will be used in 90% of tests and it gives you predefined container. You are not really interested in how it is created.

test/Microsoft.ApplicationInsights.AspNet.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
abaranch: That would overcomplicate asserts. That are not that complex.

test/Microsoft.ApplicationInsights.AspNet.Tests/Helpers/HttpContextAccessorHelper.cs
olegsych: Sure; it's the best option we have today without being able to get request-scoped services injected.

test/Microsoft.ApplicationInsights.AspNet.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
```diff
@@ -72,16 +70,107 @@ public void InitializeSetsTelemetryOperationNameToMethodAndPath()
...
+            actionContext.RouteData.Values.Add(AttributeRouting.RouteGroupKey, "RouteGroupKey");
+            
+            var contextAccessor = HttpContextAccessorHelper.CreateHttpContextAccessor(new RequestTelemetry(), actionContext);
+
+            var initializer = new OperationNameTelemetryInitializer(contextAccessor);
+
+            var telemetry = new EventTelemetry();
+            initializer.Initialize(telemetry);
+
+            Assert.Equal("GET account/login", telemetry.Context.Operation.Name);
```
olegsych: Would `Assert.DoesNotContain` be better?

abaranch:
Close without merging because I need to push to develop

test/Microsoft.ApplicationInsights.AspNet.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
olegsych: I don't quite see how replacing `Assert.Equal` with `Assert.Contains` or `Assert.EndsWith` would overcomplicate tests. :confused: Here is another question - if the asserts and tests are fine the way they are, why do we want to have 4 separate tests with mostly [duplicate code](http://xunitpatterns.com/Test%20Code%20Duplication.html) instead of just one - `InitializeSortsParameters`?

test/Microsoft.ApplicationInsights.AspNet.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
abaranch: Assert is not that complex to try to simplify it but we also check for any other side effects doing Equal

test/Microsoft.ApplicationInsights.AspNet.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
abaranch: They are not the same. You have different incoming parameters and have different resulting request name as an output. All combinations exercise different code paths. Making check less constraint does not make much sense to me.

PR #51, opened 2015-04-28
change references to point to RC packages

SergKanz:
ASP.NET runtime packages dropped build number from packages version and our package will not work with the default Visual Studio RC installation. I propose to release a hotfix today that only changes references. Going forward we can try to drop MyGet from the list of NuGet sources completely.


olegsych:
Shouldn't this merge to develop?

SergKanz:
I'm thinking to release it as a hotfix without taking all other changes now. This should definitely be ported to the develop branch as well.

olegsych:
What's the rush? Is RC out already?

abaranch:
Approved

olegsych:
:+1:

PR #54, opened 2015-04-29
measure startup time under debugger in Visual studio

SergKanz:
I run it many times - typically times very similar. Worst I get out of two sequential runs is 0.4 sec:
```
	with AI	without AI	
	6903.4364	5701.4235	
	6285.2875	5988.7779	
	6400.6323	5757.9383	
	6109.9152	6126.5687	
	6283.0474	5906.7686	
	6237.337	5944.2693	
	6199.4625	6948.1886	
	6467.1623	5916.281	
	7070.9624	5805.3368	
	6397.2912	5813.8167	overhead:
avg:	6435.45342	5990.93694	444.51648

```

### DIFF
--- a/test/StartupPerf/startupPerf.ps1
+++ b/test/StartupPerf/startupPerf.ps1
@@ -1,0 +1,47 @@
+# execute this first: Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process -Confirm 
+
+function GetUrlStatusCode($url)
+{
+	$req = [system.Net.WebRequest]::Create($url)
+	try {
+		$res = $req.GetResponse()
+	} catch [System.Net.WebException] {
+		$res = $_.Exception.Response
+	}
+	return [int]$res.StatusCode
+}
+
+$totalTime = 0
+$attempts = 10
+
+function StartUnderDebugger()
+{
+	$script:sw = [system.diagnostics.stopwatch]::StartNew()
+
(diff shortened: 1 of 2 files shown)


olegsych:
Is this startup overhead or runtime overhead?

SergKanz:
just startup under debugger. This is what ASP.NET team optimizes for and very sensitive to.

test/StartupPerf/startupPerf.ps1
```diff
@@ -0,0 +1,47 @@
...
+}
+
+$totalTime = 0
+$attempts = 10
+
+function StartUnderDebugger()
+{
+	$script:sw = [system.diagnostics.stopwatch]::StartNew()
+
+	$dte.ExecuteCommand("Debug.Start");
```
olegsych: To measure startup overhead, shouldn't the `Debug.Start` and `Debug.StopDebugging` commands be executed inside the loop? Right now, only the first GetUrlStatusCode call will incur the startup overhead, while the remaining 200 will incur only the runtime overhead.

SergKanz: Debug.Start is mostly async operation. It is like a click on menu item. After this compilation and iisepress start happens. So I need to ensure that app is actually loaded. Typically it is just couple http requests I need to do to get result.

Stop debugger is not that interesting to measure. It may be good idea though to sleep a second before the next debug session start.

Issue #25, opened 2015-04-16
Layout.cshtml: use regular extension method instead of HttpHelper


SergKanz:
Asp.Net team adviced to consider not to use @Html helper in Layout.cshtml, use extension method instead to get rid of using statement and passing InstrumentationKey explicitly

SergKanz:
[commit](https://github.com/Microsoft/ApplicationInsights-aspnetv5/commit/3af69b67fb63bc29858e56dcb31410aae5e2e168)

PR #56, opened 2015-04-30
do not use Mvc.Core for JavaScriptSnippet

SergKanz:
This addresses [issue #25](https://github.com/Microsoft/ApplicationInsights-aspnetv5/issues/25)


src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -53,41 +54,16 @@ public static void AddApplicationInsightsTelemetry(this IServiceCollection servi
                 return telemetryConfiguration;
             });
 
+            services.AddSingleton<IJavaScriptSnippet, ApplicationInsightsJavaScript>();
+
             services.AddScoped<TelemetryClient>();
 
             services.AddScoped<RequestTelemetry>((svcs) => {
```
olegsych: Is the factory method still necessary in this registration?

test/Microsoft.ApplicationInsights.AspNet.Tests/JavaScript/ApplicationInsightsJavaScriptTest.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        }
+
+        [Fact]
+        public static void SnippetWillIncludeInstrumentationKeyAsSubstring()
+        {
+            string unittestkey = "unittestkey";
+            var telemetryConfiguration = new TelemetryConfiguration { InstrumentationKey = unittestkey };
+            var snippet = new ApplicationInsightsJavaScript(telemetryConfiguration);
+            Assert.True(snippet.Write().Contains("'" + unittestkey + "'"));
```
olegsych: Assert.Contains?

test/Microsoft.ApplicationInsights.AspNet.Tests/JavaScript/ApplicationInsightsJavaScriptTest.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public static class ApplicationInsightsJavaScriptTest
+    {
+        [Fact]
+        public static void SnippetWillBeEmptyWhenInstrumentationKeyIsNotDefined()
+        {
+            var telemetryConfigurationWithEmptyKey = new TelemetryConfiguration { InstrumentationKey = string.Empty };
+            var snippet1 = new ApplicationInsightsJavaScript(telemetryConfigurationWithEmptyKey);
+            Assert.Equal(string.Empty, snippet1.Write());
+
```
olegsych: Split?

src/Microsoft.ApplicationInsights.AspNet/JavaScript/IApplicationInsightsJavaScript.cs
```diff
@@ -0,0 +1,8 @@
+﻿
+namespace Microsoft.ApplicationInsights.AspNet.JavaScript
+{
+    public interface IJavaScriptSnippet
```
olegsych: Why do we need this interface?

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
SergKanz: Yes, it tries to call not-default constructor and fails otherwise. I'll add a comment

test/Microsoft.ApplicationInsights.AspNet.Tests/JavaScript/ApplicationInsightsJavaScriptTest.cs
SergKanz: Thanks!

src/Microsoft.ApplicationInsights.AspNet/JavaScript/IApplicationInsightsJavaScript.cs
SergKanz: Idea here is that somebody may implement snippet that populates additional properties like application version or request ID. Or implement a snippet with ajax support.

BTW, what do you think of moving this functionality to JavaScript NuGet and take a reference on it?

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
olegsych: Ah, yes. I remember that quirk. Might be a good idea to leave a comment and prevent someone from "simplifying" it by accident :smile:

src/Microsoft.ApplicationInsights.AspNet/JavaScript/IApplicationInsightsJavaScript.cs
olegsych: Oh, I see. I have hard time imagining someone doing that though. It could also be done without the interface. 

We should talk to @MaxShehovtsov and @scsouthw before deciding to put this in the Microsoft.ApplicationInsights.JavaScript NuGet package. I could also imagine it being used in non-ASP.NET project types, like Cordova and WWA.

PR #57, opened 2015-04-30
Enable automated CI build of ApplicaitonInsights.AspNet.sln

olegsych:
`msbuild ApplicationInsights.AspNet.sln` fails out of the box with Visual Studio 2015 RC when any of the NuGet packages are missing from the .dnx/packages cache. This pull request adds a custom MSBuild .targets file that executes `dnu` to restore the packages. 

This pull request also adds all build-time NPM dependencies to package.json, including Bower and Gulp, to enable their automatic installation on the CI server during build.

### DIFF
--- a/ApplicationInsights.AspNet.sln
+++ b/ApplicationInsights.AspNet.sln
@@ -7,6 +7,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{2E6DDE9E-8C7
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{62AD20FD-640F-4F99-94EF-96A7581F1CF9}"
 	ProjectSection(SolutionItems) = preProject
+		Common.targets = Common.targets
 		global.json = global.json
 		NuGet.config = NuGet.config
 		Readme.md = Readme.md
--- a/Common.targets
+++ b/Common.targets
@@ -1,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="RestorePackages" Condition="'$(BuildingInsideVisualStudio)' != 'true'"
+    DependsOnTargets="GetRuntimeToolingPathTarget" BeforeTargets="CoreCompile">
+    <PropertyGroup>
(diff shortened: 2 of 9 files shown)


olegsych:
@Eilon is this the right way to implement package restore?

SergKanz:
:+1:

Eilon:
VS should restore automatically when opening the project. I don't think you should call "dnu" automatically.

Also I'm not sure I understand the NPM comment. "dnu" has no dependency on "NPM" in any way, shape, or form. If your `project.json` happens to have registered scripts that run NPM that's another story, but that's your own dependency.

davidfowl:
Just don't run it before build. It'll make your life miserable. Better to have a separate restore step in a Build.proj or in the build.cmd file.

olegsych:
@Eilon, we are working on automating our CI build. Perhaps this is a misguided idea, but since we are using the Microsoft stack, we are trying to build our solution using MSBuild with TFS 2015 Build.Preview. However, with Visual Studio 2015 RC, the `msbuild ApplicationInsights.AspNet.sln` command does not restore the missing packages. As a workaround, this pull request adds a custom MSBuild target that invokes the `dnu restore` command between the `GetRuntimeToolingPathTarget` and `CoreCompile` targets. The NPM, Bower and Gulp dependency comes from one of the functional test projects we have in our solution. DNU tries to restore NPM and Bower dependencies and then DNX runs Gulp to build the project.

@davidfowl, are you saying that running `dnu restore` before `CoreCompile` is not a good idea? The reason I wanted to do it on per-project basis is because we have projects in our repository that we don't want to build in CI and running `dnu restore` as a pre-build step for the entire repository seems like an overhead.

Eilon:
I think on the CI we usually do an explicit call to `dnu restore` before building the projects. It's not done as part of building the project itself because 99% of the time it's not needed (because the packages were just restored on the previous time I built).

olegsych:
I was able to solve the problem with Bower package restore and running Gulp on the build server. All NPM dependencies are now explicitly listed in the `package.json` of the MVC functional test project. `dnu` executes `npm install` which brings these dependencies to the local `node_modules` cache. The `project.json` has also been modified to invoke Bower and Gulp from the local cache. To solve the "path too long" error, we use the `flatten-packages` NPM module in the `postrestore` step of the `project.json` executed by the `dnu`. 

Installing Bower and Gulp locally is slow, especially because of `flatten-packages`. We can avoid doing that by running `npm install -g bower` and `npm install -g gulp` commands on the build agent and making sure that NPM's global package cache `%USERPROFILE%\AppData\Roaming\npm` has to be added to `%PATH%`.

PR #64, opened 2015-05-05
implement SDK telemetry for ASP.NET 5 applications

SergKanz:


src/Microsoft.ApplicationInsights.AspNet/SdkVersionUtils.cs
```diff
@@ -0,0 +1,17 @@
...
+    using System.Reflection;
+
+    internal class SdkVersionUtils
+    {
+        internal const string VersionPrefix = "aspnet5: ";
+
+        internal static string GetAssemblyVersion()
+        {
+            return typeof(SdkVersionUtils).GetTypeInfo().Assembly.GetCustomAttributes<AssemblyFileVersionAttribute>()
+                    .First()
```
abaranch: Is it always available?

abaranch:
Approved

src/Microsoft.ApplicationInsights.AspNet/SdkVersionUtils.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Microsoft.ApplicationInsights.AspNet
+{ 
+    using System.Linq;
+    using System.Reflection;
+
+    internal class SdkVersionUtils
```
olegsych: Can we convert this class to an `IContextInitializer` or an `ITelemetryInitializer`?

src/Microsoft.ApplicationInsights.AspNet/SdkVersionUtils.cs
olegsych: (Reflection + Linq) x 2 middlewares per request. :scream: Can we cache this value?

test/Microsoft.ApplicationInsights.AspNet.Tests/RequestTrackingMiddlewareTest.cs
```diff
@@ -0,0 +1,43 @@
...
+            };
+            var middleware = new RequestTrackingMiddleware(nextMiddleware, MockTelemetryClient());
+
+            await middleware.Invoke(new DefaultHttpContext(), new RequestTelemetry());
+
+            Assert.NotEmpty(sentTelemetry.Context.GetInternalContext().SdkVersion);
+            Assert.Contains("aspnetv5", sentTelemetry.Context.GetInternalContext().SdkVersion);
+        }
+
+        private TelemetryClient MockTelemetryClient()
```
olegsych: This looks similar to the helper method in the `ExceptionTrackingMiddlewareTest`. Can we share it?

src/Microsoft.ApplicationInsights.AspNet/ExceptionTrackingMiddleware.cs
```diff
@@ -13,11 +14,13 @@ public sealed class ExceptionTrackingMiddleware
     {
         private readonly RequestDelegate next;
         private readonly TelemetryClient telemetryClient;
+        private readonly string sdkVersion;
 
         public ExceptionTrackingMiddleware(RequestDelegate next, TelemetryClient client)
         {
             this.next = next;
             this.telemetryClient = client;
+            this.sdkVersion = SdkVersionUtils.VersionPrefix + SdkVersionUtils.GetAssemblyVersion();
```
olegsych: Why are the middleware components responsible for setting the SdkVersion?

src/Microsoft.ApplicationInsights.AspNet/RequestTrackingMiddleware.cs
```diff
@@ -16,14 +19,16 @@ public sealed class RequestTrackingMiddleware
...
         
         public RequestTrackingMiddleware(RequestDelegate next, TelemetryClient client)
         {
             this.telemetryClient = client;
             this.next = next;
+            this.sdkVersion = SdkVersionUtils.VersionPrefix + SdkVersionUtils.GetAssemblyVersion();
         }
 
-        public async Task Invoke(HttpContext httpContext)
+        public async Task Invoke(HttpContext httpContext, RequestTelemetry telemetry)
```
olegsych: :+1:

src/Microsoft.ApplicationInsights.AspNet/SdkVersionUtils.cs
SergKanz: We are using the same code in Core

src/Microsoft.ApplicationInsights.AspNet/SdkVersionUtils.cs
SergKanz: The idea here is only mark auto-generated data with aspnet5 keyword, not everything

test/Microsoft.ApplicationInsights.AspNet.Tests/RequestTrackingMiddlewareTest.cs
SergKanz: makes sense. I guess we will have more usage of this helper method

src/Microsoft.ApplicationInsights.AspNet/SdkVersionUtils.cs
olegsych: Ah, I remember now. It's the approach neither of us liked :smile:. Why are we doing it again? Does this require changes in our telemetry reports?

src/Microsoft.ApplicationInsights.AspNet/ExceptionTrackingMiddleware.cs
SergKanz: The idea here is to count how many people are using auto-collection, not making TrackException manually

src/Microsoft.ApplicationInsights.AspNet/SdkVersionUtils.cs
SergKanz: I just believe in this pattern. What can I do to convince that it is good? Imagine - when you doing TrackMetric you can distinct whether it is call to TrackMetric or PerformanceCountersCollection module or @lukekim 's metric aggregation module. Will it be cool?

olegsych: I understand what we are trying to achieve here (thanks for the reminder :smile:). However, putting this information in the SdkVersion field violates the [single responsibility principle](http://www.butunclebob.com/ArticleS.UncleBob.PrinciplesOfOod) of OO and the [first normal form](http://en.wikipedia.org/wiki/First_normal_form) of Database design, which leads to known consequences. Will the data pipeline separate the "source" information into a different field in storage schema for filtering and segmentation? Are we working with them to introduce a new field for this information in the data collection schema?

Issue #39, opened 2015-04-21
Failed request status code is 200


abaranch:
Update: issue summary:

- If exception wasn't handled by any middleware runtime will still return 500. Application Insights will report 200.
- MusicStore application is using [UseStatusCodePagesWithRedirects](https://github.com/aspnet/Diagnostics/blob/b1643b438aa947370868b4d5ee7727c27f2d78cb/src/Microsoft.AspNet.Diagnostics/StatusCodePagesExtensions.cs#L76) which hiding error response code from customer and Application Insights. So technically Application Insights returns correct status code, however it will never show the number of "Failed requests" increasing.

Question here - what kind of heuristic can we use (or should we at all?) to report the proper response code.

olegsych:
I could not repro this problem. After discussing with @abaranch, it may have occurred if the Application Insights `RequestTrackingMiddleware` was added after the ASP.NET's [ErrorHandlerMiddleware](https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNet.Diagnostics/ErrorHandlerMiddleware.cs), which is responsible for setting the `HttpResponse.StatusCode`. When added correctly, as the first middleware in `Configure` method, the `RequestTrackingMiddleware` correctly sets `RequestTelemetry.StatusCode` from `HttpResponse.StatusCode`.

olegsych:
It turns out that the ErrorPageMiddleware, used by aspnet in development mode does not set the status code.

olegsych:
Opened aspnet/diagnostics#133 to understand how the `Response.StatusCode` is meant to be set.

olegsych:
Opened aspnet/Diagnostics#133 to understand how `Response.StatusCode` is supposed to be set.

SergKanz:
@Eilon I updated description of the issue. Who will be the best person to discuss error processing practices in ASP.NET 5 and how Application Insights should implement it to show credible data to developer?

Eilon:
Logging would get you more info. But ultimately there's obviously no absolute way to get all possible information because components are free to have undiagnosable behavior. Having said that, the built-in components should "play well with others" as much as possible.

I'm not sure how you could have an HTTP 500 (or whatever) to represent an error yet also redirect to another page... maybe it's a best practice not to do redirects on errors to begin with.

SergKanz:
[...]

PR #66, opened 2015-05-07
exceptions functional tests

SergKanz:
Basic validation of exceptions in standard application templates.

@Eilon - standard Web API application template doesn't have error handling and will return 500 when controller throws unhandled exception. Should we file an bug or it is a known issue?


test/FunctionalTestUtils/Tests/RequestTelemetryTestsBase.cs
```diff
@@ -18,8 +19,9 @@ public void ValidateBasicRequest(string requestPath, RequestTelemetry expected)
             task.Wait(TestTimeoutMs);
             var result = task.Result;
             
-            Assert.Equal(1, this.Buffer.Count);
-            ITelemetry telemetry = this.Buffer[0];
+            var items = this.Buffer.Where((item) => { return item is RequestTelemetry; });
```
olegsych: Can we use `.OfType<RequestTelemetry>()` instead of `.Where`?

test/FunctionalTestUtils/Tests/RequestTelemetryTestsBase.cs
```diff
@@ -31,5 +33,25 @@ public void ValidateBasicRequest(string requestPath, RequestTelemetry expected)
             Assert.True(actual.Duration < (DateTimeOffset.Now - testStart), "duration");
             Assert.Equal(expected.HttpMethod, actual.HttpMethod);
         }
+
+        public void ValidateBasicException(string requestPath, ExceptionTelemetry expected)
```
olegsych: Is this test in the right class? It verifies `ExceptionTelemetry` instances while the class seems to imply its for `RequestTelemetry` instances. Alternatively, is this base class needed? Can these two methods be pulled up or pushed down?

test/FunctionalTestUtils/Tests/RequestTelemetryTestsBase.cs
```diff
@@ -31,5 +33,25 @@ public void ValidateBasicRequest(string requestPath, RequestTelemetry expected)
...
+            var task = this.HttpClient.GetAsync(this.Server.BaseHost + requestPath);
+            task.Wait(TestTimeoutMs);
+            var result = task.Result;
+
+            var items = this.Buffer.Where((item) => { return item is ExceptionTelemetry; });
+            Assert.Equal(1, items.Count());
+            ITelemetry telemetry = items.First();
+            Assert.IsAssignableFrom(typeof(ExceptionTelemetry), telemetry);
+            var actual = (ExceptionTelemetry)telemetry;
+
```
olegsych: Sorry, my [OCD](http://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_disorder) compels me to say this :blush:  With Linq and xUnit, this code block could be shortened to 
```c#
ITelemetry telemetry = this.Buffer.OfType<ExceptionTelemetry>().Single();
var actual = Assert.IsType<ExceptionTelemetry>(telemetry);
```

SergKanz:
I want to make an observation here - when controller throws an exception and there is no error handling - server will return 500. Now the question - should our request middleware assume 500 if it sees exception?

test/WebApiShimFw46.FunctionalTests/Controllers/ValuesController.cs
```diff
@@ -20,6 +20,10 @@ public IEnumerable<string> Get()
         [HttpGet("{id}")]
         public string Get(int id)
         {
+            if (id == 42)
```
olegsych: This is a very profound id value :grinning:. Existential issues aside, I think tests would be cleaner with a separate method, similar to the `Exception` method in the MVC functional test.

olegsych:
What if we assume it's 500 and another middleware further up the stack decides its 400 or something else?

SergKanz:
Exactly. @Eilon, who is the best person to discuss how Application Insights should behave in this situation?

Eilon:
I saw the bugs you opened in Diagnostics, in particular, https://github.com/aspnet/Diagnostics/issues/133. I think that bug is a valid bug, but adding @Tratcher to comment to see if he concurs.

As far as how best to deal with composition and other middleware changing stuff from underneath you: I don't think there's any way to guarantee this - the system is basically designed to prevent that from being possible. If intermediate middleware changes the status code or eats an exception, you'll never see it, but that's completely by design.

SergKanz:
@Eilon, we found two cases Application Insights is now working as one would expect:
- If exception wasn't handled by anybody runtime will still return 500
- If request wasn't processed at all - runtime will return 404
Question here - what kind of heuristic can we use (or should we at all?) to report the proper response code.

SergKanz:
We have an issue #39 to track this problem. It wasn't wise to start conversation in pull request.

PR #70, opened 2015-05-08
use htmlhelpter in cshtml

SergKanz:
Addressing @Eilon feedback on integration points

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsJavaScript.cs
+++ b/src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsJavaScript.cs
@@ -1,8 +1,10 @@
 ﻿
-namespace Microsoft.ApplicationInsights.AspNet.JavaScript
+namespace Microsoft.ApplicationInsights.AspNet
 {
     using System;
     using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.AspNet.Mvc.Rendering;
+
 
     /// <summary>
     /// This class helps to inject Application Insights JavaScript snippet into applicaiton code.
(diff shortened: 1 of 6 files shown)


abaranch:
Approved

src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsJavaScript.cs
```diff
@@ -13,7 +15,7 @@ namespace Microsoft.ApplicationInsights.AspNet.JavaScript
     ///         @Html.Raw(ApplicationInsightsJavaScript.Write())
```
upendras: nit - we ned to update the comment

upendras:
:shipit:

src/Microsoft.ApplicationInsights.AspNet/ApplicationInsightsJavaScript.cs
```diff
@@ -1,19 +1,15 @@
...
-    /// Usage:
-    ///     Inject an instance of the class into your razor code using code:
-    ///         @inject Microsoft.ApplicationInsights.AspNet.JavaScript.IJavaScriptSnippet ApplicationInsightsJavaScript
-    ///     Place this code fragment just before closing "head" tag
-    ///         @Html.Raw(ApplicationInsightsJavaScript.Write())
-    ///     &lt;/head&gt;
     /// </summary>
-    public class ApplicationInsightsJavaScript : IJavaScriptSnippet
+    public class JavaScriptSnippet
     {
```
olegsych: So at this point, `JavaScriptSnippet` is an implementation detail of the `ApplicationInsightsJavaScript` extension method and does not need to be public. Correct? Is that something we care about?

Issue #69, opened 2015-05-08
Reference ASP.NET "Abstractions" / "Interfaces" packages instead of concrete implementations


Eilon:
Many of the packages in ASP.NET 5 and DNX now have "Abstractions" packages (as of beta5; in beta4 these were generally "Interfaces" packages) that contain just the raw interfaces, without specific implementations. As a general rule, library packages (such as Application Insights) should reference the abstractions and not concrete implementations.

E.g. the references here: https://github.com/Microsoft/ApplicationInsights-aspnet5/blob/master/src/Microsoft.ApplicationInsights.AspNet/project.json#L13-L17

You should be able to change the reference to `Microsoft.Framework.Logging` to instead be `Microsoft.Framework.Logging.Abstractions` (in beta5; or use `Microsoft.Framework.Logging.Interfaces` in beta4).

SergKanz:
Fixed in [PR 72](https://github.com/Microsoft/ApplicationInsights-aspnet5/pull/72)

PR #72, opened 2015-05-08
use Interfaces references instead of real pakages

SergKanz:
Address #69


abaranch:
Approved

Eilon:
Is it possible to change the reference to `Microsoft.AspNet.Mvc.Core` to be `Microsoft.AspNet.Mvc.Abstractions`? More info here: https://github.com/aspnet/Announcements/issues/16

@rynowak to see if he has thoughts on that.

rynowak:
There's no `.Mvc.Abstractions` in beta4 - also looks like this code https://github.com/Microsoft/ApplicationInsights-aspnet5/blob/master/src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsJavaScriptExtensions.cs#L14 defines an extension method for `IHtmlHelper` so `.Abstractions` is a no-go for now.

PR #73, opened 2015-05-09
make functional tests pass all the time

SergKanz:
Three issues:

1. Unit tests run in parallel in Visual Studio. So Random returns the same value quite often. Make it static.
2. Made telemetry channel non-singleton 
3. Time calculation was very un-precise when comparing with duration value of RequestTelemetry. Use stopwatch instead.


abaranch:
Approved

PR #80, opened 2015-05-14
mark request as failed for runaway exceptions

SergKanz:


src/Microsoft.ApplicationInsights.AspNet/RequestTrackingMiddleware.cs
```diff
@@ -32,31 +29,33 @@ public RequestTrackingMiddleware(RequestDelegate next, TelemetryClient client)
...
 
-                    telemetry.Timestamp = now;
-                    telemetry.Duration = sw.Elapsed;
-                    telemetry.ResponseCode = httpContext.Response.StatusCode.ToString();
-                    telemetry.Success = httpContext.Response.StatusCode < 400;
-                    telemetry.HttpMethod = httpContext.Request.Method;
-                    telemetry.Url = httpContext.Request.GetUri();
-                    telemetry.Context.GetInternalContext().SdkVersion = this.sdkVersion;
+                telemetry.Duration = sw.Elapsed;
+                telemetry.ResponseCode = httpContext.Response.StatusCode.ToString();
```
abaranch: So it still will be 200. Are we planning to di smth with that?..

abaranch:
Approved

src/Microsoft.ApplicationInsights.AspNet/RequestTrackingMiddleware.cs
```diff
@@ -32,31 +29,33 @@ public RequestTrackingMiddleware(RequestDelegate next, TelemetryClient client)
...
-            var now = DateTimeOffset.UtcNow;
+            bool requestFailed = false;
 
             try
             {
                 await this.next.Invoke(httpContext);
             }
+            catch (Exception)
+            {
+                requestFailed = true;
```
upendras: should we rethrow this exception?

SergKanz: Great catch! Thanks!

PR #92, opened 2015-06-04
Fix unit tests in CI

upendras:
- Use test project directory as the current working directory during
execution.
- Figured out issue#91 and updated test expectation for the failing test
for now.
- ASP.Net Error handling middleware logs failed requests as Error and
the ConsoleLogger ends up writing it to powershell resulting in CI
failing. Moved ConsoleLogger to only Development environment.


Sergei-Nikitin:
:shipit:

PR #109, opened 2015-08-19
Changes for beta7

upendras:


abaranch:
:shipit:

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/OperationNameTelemetryInitializer.cs
```diff
@@ -9,82 +9,98 @@
...
+        public void OnBeforeAction(ActionDescriptor actionDescriptor, HttpContext httpContext, RouteData routeData)
+        {
+            string name = this.GetNameFromRouteContext(routeData);
+            var telemetry = httpContext.RequestServices.GetService<RequestTelemetry>();
 
-                telemetry.Context.Operation.Name = name;
+            if (!string.IsNullOrEmpty(name) && telemetry != null)
+            {
+                name = httpContext.Request.Method + " " + name;
+                ((RequestTelemetry)telemetry).Name = name;
```
SergeyKanzhelev: check that telemetry is in fact RequestTelemetry to avoid NullReferenceException

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/OperationNameTelemetryInitializer.cs
```diff
@@ -9,82 +9,98 @@
...
+                ((RequestTelemetry)telemetry).Context.Operation.Name = name;
             }
```
SergeyKanzhelev: let's use URL otherwise

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/OperationNameTelemetryInitializer.cs
```diff
@@ -9,82 +9,98 @@
...
-                
-                var telemetryType = telemetry as RequestTelemetry;
-                if (telemetryType != null && string.IsNullOrEmpty(telemetryType.Name))
+                else
                 {
-                    telemetryType.Name = name;
+                    // We didn't get BeforeAction notification
+                    string name = platformContext.Request.Method + " " + platformContext.Request.Path.Value;
+                    requestTelemetry.Name = name;
+                    requestTelemetry.Context.Operation.Name = name;
```
SergeyKanzhelev: no need to set it - it will be set by the same telemetry initializer when it will be called for RequestTelemetry

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/OperationNameTelemetryInitializer.cs
```diff
@@ -9,82 +9,98 @@
...
+        {
+            string name = this.GetNameFromRouteContext(routeData);
+            var telemetry = httpContext.RequestServices.GetService<RequestTelemetry>();
 
-                telemetry.Context.Operation.Name = name;
+            if (!string.IsNullOrEmpty(name) && telemetry != null)
+            {
+                name = httpContext.Request.Method + " " + name;
+                ((RequestTelemetry)telemetry).Name = name;
+                ((RequestTelemetry)telemetry).Context.Operation.Name = name;
```
SergeyKanzhelev: no need to set it - it will be set by the same telemetry initializer when it will be called for RequestTelemetry

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/OperationNameTelemetryInitializer.cs
upendras: that is taken care of in OnInitializeTelemetry() method when RequestTelemetry.Name is found null or empty. We expect that if we reveive a notification from MVC, it has proper RouteData.

PR #111, opened 2015-09-02
Merge develop to master

upendras:


src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/OperationNameTelemetryInitializer.cs
```diff
@@ -9,82 +9,96 @@
...
             }
         }
 
-        private string GetNameFromRouteContext(IServiceProvider requestServices)
+        private string GetNameFromRouteContext(RouteData routeData)
         {
             string name = null;
 
-            if (requestServices != null)
+            if (routeData.Values.Count > 0)
```
abaranch: Can that be null?

abaranch: :shipit:

upendras: will confirm.

PR #113, opened 2015-09-17
allow monitoring of non-MVC apps

SergeyKanzhelev:
Fixing https://github.com/Microsoft/ApplicationInsights-aspnet5/issues/112


upendras:
:shipit:

upendras:
I think there is a unit test that expects the initializer to throw on passing null, lets remove that as well.

SergeyKanzhelev:
I deleted this test. See the last file in PR

PR #123, opened 2016-01-14
Updating AI version from 1.0.0 to 2.0.0.beta4

kartang:


NuGet.config
```diff
@@ -3,7 +3,8 @@
   <packageSources>
     <add key="ASP.NET 5 Dev" value="https://www.myget.org/F/aspnetvnext/" />
     <add key="ASP.NET 5" value="https://www.myget.org/F/aspnetrelease/" />
-    <add key="nuget.org feed" value="https://www.nuget.org/api/v2/" />
+    <add key="nuget.org feed" value="https://www.nuget.org/api/v2/" />    
+    <add key="UpdatedVersion" value="C:\Users\kartang\Documents\GitHub\NugetPackages" />
   </packageSources>
```
kartang: We should remove this, but I am not sure if we need to point to the actual drop folder.

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/DomainNameRoleInstanceTelemetryInitializer.cs
```diff
@@ -5,31 +5,25 @@
...
     using System.Threading;
-
+    using Channel;
     using Microsoft.ApplicationInsights.DataContracts;
-    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.AspNet.Http;
+    using TelemetryInitializers;
 
-    /// <summary>
-    /// A telemetry context initializer that populates device context role instance.
```
SergeyKanzhelev: have you removed comment completely? Please return back if you did - it is a public class

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/DomainNameRoleInstanceTelemetryInitializer.cs
```diff
@@ -5,31 +5,25 @@
...
+        }
 
-            if (string.IsNullOrEmpty(context.Device.RoleInstance))
+        protected override void OnInitializeTelemetry(HttpContext platformContext, RequestTelemetry requestTelemetry, ITelemetry telemetry)
+        {
+            if (string.IsNullOrEmpty(telemetry.Context.Device.RoleInstance))
             {
                 var name = LazyInitializer.EnsureInitialized(ref this.roleInstanceName, this.GetMachineName);
-                context.Device.RoleInstance = name;
+                telemetry.Context.Device.RoleInstance = name;
```
SergeyKanzhelev: Use `telemetry.Context.Cloud.RoleInstance` instead

NuGet.config
```diff
@@ -3,7 +3,8 @@
   <packageSources>
     <add key="ASP.NET 5 Dev" value="https://www.myget.org/F/aspnetvnext/" />
     <add key="ASP.NET 5" value="https://www.myget.org/F/aspnetrelease/" />
-    <add key="nuget.org feed" value="https://www.nuget.org/api/v2/" />
+    <add key="nuget.org feed" value="https://www.nuget.org/api/v2/" />    
+    <add key="UpdatedVersion" value="C:\Users\kartang\Documents\GitHub\NugetPackages" />
```
SergeyKanzhelev: Use myget for temporary versions. You can use local folder during testing

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/DomainNameRoleInstanceTelemetryInitializer.cs
kartang: Added it!

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/DomainNameRoleInstanceTelemetryInitializer.cs
```diff
@@ -5,31 +5,29 @@
...
-    /// A telemetry context initializer that populates device context role instance.
-    /// </summary>
-    public class DomainNameRoleInstanceContextInitializer : IContextInitializer
+    public class DomainNameRoleInstanceTelemetryInitializer : TelemetryInitializerBase
     {
         private string roleInstanceName;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="DomainNameRoleInstanceContextInitializer"/> class.
+        /// A telemetry initializer that populates device context role instance.
```
SergeyKanzhelev: this should be a class comment, not a method comment

SergeyKanzhelev:
Fix comment and :shipit:

AlexBulankou:
Why are we using beta4 and not 1.2.3 that has all the required fixes for vs output? I would think for these changes we would like to go with the least risky approach which would be the stable 1.2.3.

SergeyKanzhelev:
@AlexBulankou  One of the changes we need is to support generations - new .NET moniker. Thus the change.

Another consideration - we really hope to get stable 2.0 by the time of ASP.NET 5 release. So it will look bad to still have a 1.2.3 reference in ASP.NET 5

PR #128, opened 2016-01-29
Kartang/rdd test

kartang:


test/FunctionalTestUtils/project.json
```diff
@@ -4,21 +4,25 @@
...
-        "Microsoft.ApplicationInsights": "2.0.0-beta4",
-        "Microsoft.ApplicationInsights.AspNet": "1.0.0-rc1",
-        "Microsoft.AspNet.Http.Abstractions": "1.0.0-rc1-*",
-        "Microsoft.AspNet.Server.WebListener": "1.0.0-rc1-*",
-        "Microsoft.Dnx.Runtime": "1.0.0-rc1-*",
-        "xunit.runner.aspnet": "2.0.0-aspnet-rc1-*",
-        "System.Net.Http": "4.0.0-beta-23109"
-    },
+  "dependencies": {
+    "Microsoft.ApplicationInsights": "2.0.0-beta4",
```
kartang: Will change beta4 to * as soon as I remove the myget, in the next iteration.

test/FunctionalTestUtils/BackTelemetryChannel.cs
```diff
@@ -37,12 +37,12 @@ public string EndpointAddress
         {
             get
             {
-                throw new NotImplementedException();
```
SergeyKanzhelev: create an issue on server-dotnet repo to not fail dependency collection when channel misbehaving

test/FunctionalTestUtils/TelemetryTestsBase.cs
```diff
@@ -48,5 +52,42 @@ public void ValidateBasicException(InProcessServer server, string requestPath, E
             Assert.NotEmpty(actual.Context.Operation.Name);
             Assert.NotEmpty(actual.Context.Operation.Id);
         }
+
+        public void ValidateBasicDependency(InProcessServer server, string requestPath, DependencyTelemetry expected)
+        {
+
+#if dnx451
```
SergeyKanzhelev: Let's do this if around entire method so we will not see passed dependency test for dnxcore

SergeyKanzhelev:
thanks for cleaning up test file names

SergeyKanzhelev:
:shipit:

test/FunctionalTestUtils/TelemetryTestsBase.cs
kartang: The actual tests are completely wrapped around the condition (which will not account to the passed count). This is a function that is being used by the tests. It also makes sense to wrap around this. Will do that.

PR #135, opened 2016-02-08
Use rc1-final instead of rc1-update1

SergeyKanzhelev:


AlexBulankou:
@kartang can you please take a look? Thanks!

PR #142, opened 2016-03-17
Update to RC2, NetStandard and CLI

kartang:
Changes:
1) Update the dependencies to RC2
2) Change the framework runtime to use CLI from DNX
3) Change the framework moniker from dnx451->net451 and dnxcore50->netstandard1.3
4) Change the package name to ApplicationInsights.AspNetCore from ApplicationInsights.AspNet
5) Update the tests accordingly.

Issues/Questions:
1) Tests from EmptyApp.FunctionalTests and Microsoft.ApplicationInsights.AspnetCore.Tests run fine. However, we are having trouble running the functional tests for MVC and WEBAPI projects. 

**Note: Added some comments to important changes.**


kartang:
@davidfowl / @anurse : Can you please take a look at the transition from dnxcore50->netstandard1.3 and dnx to CLI and let us know if that looks OK (for Microsoft.ApplicationInsights.AspnetCore project). Futhermore, Can you please let us know if we need to  configure anything for MVC test application. The application as of now is always returning 404's for all the pages. Thank you!

global.json
```diff
@@ -1,8 +1,3 @@
 ﻿{
-    "sources": [ "src", "test" ],
-    "sdk": {
-            "architecture": "x86",
-      "runtime": "clr",
```
kartang: Removed these as a part of CLI transition.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
...
+      "compilationOptions": {
+        "define": [ "net451" ]
+      },
+      "dependencies": {
+        "Microsoft.ApplicationInsights.PerfCounterCollector": "2.0.0",
+        "Microsoft.ApplicationInsights.DependencyCollector": "2.0.0"
+      }
+    },
+    "netstandard1.3": {
+      "imports": [
```
kartang: dnx451->net451 and dnxcore50->netstandard1.3 
Microsoft.ApplicationInsigts has pcl that supports framework "dotnet5.4", which is why I added it in the imports below. Please let me know if it is required forever.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/project.json
```diff
@@ -0,0 +1,41 @@
...
+  "frameworks": {
+    "netstandardapp1.5": {
+      "imports": [
+        "dnxcore50",
+        "dotnet5.4",
+        "portable-net451+win8"
+      ],
+      "dependencies": {
+        "moq.netcore": "4.4.0-beta8",
+        "NETStandard.Library": "1.0.0-*",
```
kartang: Removed net451 as I had some trouble running tests for that framework. We will add it back when we have a stable CLI version, and that is the reason why we have the dependencies left behind under netstandardapp!

test/WebApiShimFw46.FunctionalTests/WebApiShimFw46.FunctionalTests.xproj
```diff
@@ -4,7 +4,7 @@
     <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
   </PropertyGroup>
-  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.Props" Condition="'$(VSToolsPath)' != ''" />
+  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''" />
   <PropertyGroup Label="Globals">
```
kartang: Part of DNX to CLI conversion

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/DomainNameRoleInstanceTelemetryInitializer.cs
```diff
@@ -35,7 +35,7 @@ private string GetMachineName()
             string hostName = Dns.GetHostName();
 
             // Issue #61: For dnxcore machine name does not have domain name like in full framework 
-#if !dnxcore50
+#if !NETSTANDARD1_3
```
kartang: I believe this is automatically generated by CLI.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/WebUserTelemetryInitializerTests.cs
```diff
@@ -56,37 +54,12 @@ public void InitializeDoesNotOverrideUserProvidedInline()
...
-            var requestTelemetry = new RequestTelemetry();
-            var contextAccessor = HttpContextAccessorHelper.CreateHttpContextAccessor(requestTelemetry);
             contextAccessor.HttpContext.Request.Headers["Cookie"] = "ai_user=test";
             var initializer = new WebUserTelemetryInitializer(contextAccessor);
 
             initializer.Initialize(requestTelemetry);
 
-            Assert.Equal(null, requestTelemetry.Context.User.Id);
-        }
-
```
kartang: Removed tests as they are not required anymore (removal of acquisition date).

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
+﻿{
+  "title": "Application Insights for ASP.NET Core Web Applications",
+  "description": "Application Insights for ASP.NET Core web applications. See http://go.microsoft.com/fwlink/?LinkID=510752 for more information.",
```
AlexBulankou: I think this is the correct link to point them to: https://azure.microsoft.com/en-us/documentation/articles/app-insights-asp-net-five/

kartang: Sure, will update that!

global.json
kartang: Seems like CLI is creating WIN7-x64 build configuration by default (verified it with a sample application). Please let me know if it can be configured from here.

NuGet.config
```diff
@@ -3,7 +3,13 @@
   <packageSources>
     <add key="ASP.NET 5 Dev" value="https://www.myget.org/F/aspnetvnext/" />
     <add key="ASP.NET 5" value="https://www.myget.org/F/aspnetrelease/" />
-    <add key="nuget.org feed" value="https://www.nuget.org/api/v2/" />    
+    <add key="nuget.org feed" value="https://www.nuget.org/api/v2/" />
+    <add key="dotnet.myget.org dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
```
dnduffy: Doesn't v3 of NuGet (nuget.org) replace v2?  Why do we have both links?

src/Microsoft.ApplicationInsights.AspNetCore/Properties/launchSettings.json
```diff
@@ -0,0 +1,19 @@
+{
+  "iisSettings": {
+    "windowsAuthentication": false,
+    "anonymousAuthentication": true,
+    "iisExpress": {
+      "applicationUrl": "http://localhost:4367/",
+      "sslPort": 0
```
dnduffy: Does 0 mean no SSL?  Or should this be 443?

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Helpers/HttpContextAccessorHelper.cs
```diff
@@ -22,18 +22,18 @@ public static HttpContextAccessor CreateHttpContextAccessor(RequestTelemetry req
...
 
-            services.AddInstance<IHttpContextAccessor>(contextAccessor);
+            services.AddSingleton<IHttpContextAccessor>(contextAccessor);
 
             if (actionContext != null)
             {
                 var si = new ActionContextAccessor();
                 si.ActionContext = actionContext;
-                services.AddInstance<IActionContextAccessor>(si);
+                services.AddSingleton<IActionContextAccessor>(si);
```
dnduffy: Should context stuff be singleton or scope?

test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/DomainNameRoleInstanceTelemetryInitializerTests.cs
```diff
@@ -17,7 +17,7 @@ public class DomainNameRoleInstanceTelemetryInitializerTests
         [Fact]
         public void InitializeThrowIfHttpContextAccessorIsNull()
         {
-            Assert.Throws<ArgumentNullException>(() =>
+           var ex = Assert.ThrowsAny<ArgumentNullException>(() =>
```
dnduffy: Why grab a reference to the exception if it is not then used later?  I'm guessing it is an exception due to the lack of readability from using var.

NuGet.config
davidfowl: Both shouldn't be there

test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/DomainNameRoleInstanceTelemetryInitializerTests.cs
kartang: That was accidentally placed while debugging. I will remove that.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Helpers/HttpContextAccessorHelper.cs
kartang: It didn't really matter here, as we update the context on every test. We update the service associated with request telemetry in the context on every run, which as a result will not effect if the service is singleton or scoped.

NuGet.config
kartang: Resolved! Removed the nuget feeds and added the corresponding AI myget feed!

test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/DomainNameRoleInstanceTelemetryInitializerTests.cs
kartang: resolved!

test/Mvc6Framework45.FunctionalTests/Startup.cs
```diff
@@ -107,7 +106,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
             // Add Application Insights exceptions handling to the request pipeline.
             app.UseApplicationInsightsExceptionTelemetry();
             // Add static files to the request pipeline.
-            app.UseStaticFiles();
+            app.UseStaticFiles();
```
dnduffy: There are extra trailing spaces on the line now.

dnduffy:
:)

src/Microsoft.ApplicationInsights.AspNetCore/project.json
davidfowl: It doesn't persist into the package so thats fine but you wont need it with the latest packages

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
...
+  "tags": [ "Analytics", "ApplicationInsights", "Telemetry", "AppInsights", "ASP.NET Core" ],
+  "compilationOptions": {
+    "keyFile": "../../keys/35MSSharedLib1024.snk",
+    "delaySign": true
+  },
+  "dependencies": {
+    "Microsoft.AspNetCore.Hosting.Abstractions": "1.0.0-rc2-*",
+    "Microsoft.AspNetCore.Http.Abstractions": "1.0.0-rc2-*",
+    "Microsoft.AspNetCore.Http.Features": "1.0.0-rc2-*",
+    "Microsoft.AspNetCore.Mvc.ViewFeatures": "1.0.0-*",
```
davidfowl: What do you use from ViewFeatures? 

/cc @rynowak

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
...
+    "Microsoft.Extensions.Logging.Abstractions": "1.0.0-rc2-*",
+    "Microsoft.Extensions.Configuration": "1.0.0-rc2-*",
```
davidfowl: Which config providers do you use? Should this be abstractions?

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
...
+  "iconUrl": "http://appanacdn.blob.core.windows.net/cdn/icons/aic.png",
+  "tags": [ "Analytics", "ApplicationInsights", "Telemetry", "AppInsights", "ASP.NET Core" ],
+  "compilationOptions": {
+    "keyFile": "../../keys/35MSSharedLib1024.snk",
+    "delaySign": true
+  },
+  "dependencies": {
+    "Microsoft.AspNetCore.Hosting.Abstractions": "1.0.0-rc2-*",
+    "Microsoft.AspNetCore.Http.Abstractions": "1.0.0-rc2-*",
+    "Microsoft.AspNetCore.Http.Features": "1.0.0-rc2-*",
```
davidfowl: Abstractions depends on features, why do you need both?

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
...
+      },
+      "dependencies": {
+        "Microsoft.ApplicationInsights.PerfCounterCollector": "2.0.0",
+        "Microsoft.ApplicationInsights.DependencyCollector": "2.0.0"
+      }
+    },
+    "netstandard1.3": {
+      "imports": [
+        "dotnet5.4",
+        "dnxcore50",
```
davidfowl: You should remove dnxcore50.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
...
+        "portable-net45+win8"
+      ],
+      "dependencies": {
+        "NETStandard.Library": "1.0.0-rc2-*",
```
davidfowl: We're still working on guidance for when to use this package. I'd suggest avoiding it for now.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
...
+        "System.Net.NameResolution": "4.0.0-beta-23302",
```
davidfowl: why are you using wildcards for all other versions but a hardcoded beta version for this package?

test/EmptyApp.FunctionalTests/project.json
```diff
@@ -1,31 +1,37 @@
...
-    "xunit": "2.1.0-rc1-*",
-    "Microsoft.AspNet.Diagnostics": "1.0.0-rc1-final"
+    "Microsoft.AspNetCore.Server.WebListener": "0.1.0-rc2-*",
+    "Microsoft.AspNetCore.Diagnostics": "1.0.0-rc2-*",
+    "Microsoft.ApplicationInsights.AspNetCore": "1.0.0-rc2-*",
+    "FunctionalTestUtils": "1.0.0-rc2"
   },
-
+  "testRunner": "xunit",
   "commands": {
```
davidfowl: If you're using the CLI these aren't useful.

test/FunctionalTestUtils/InProcessServer.cs
```diff
@@ -53,14 +52,15 @@ private BackTelemetryChannel Start(string assemblyName)
             configBuilder.Add(customConfig);
             var config = configBuilder.Build();
 
-            var engine = CreateBuilder(config)
-                .UseServer("Microsoft.AspNet.Server.WebListener")
+            this.hostingEngine = CreateBuilder(config)
```
davidfowl: There's a much cleaner way to set the URL in the new hosting API. You can get rid of the config goop.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
kartang: We use htmlString and Ihtmlhelper from ViewFeatures.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
kartang: Removed features. Will be updated in the next commit.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
kartang: We use MemoryConfigurationProvider, and therefore, use Configuration as opposed to Configuration.Abstractions. Please let me know if this is OK.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
kartang: Removed it! Will push the changes in the next commit.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
kartang: Removed it for now.

test/EmptyApp.FunctionalTests/project.json
kartang: Removed the commands.

test/Mvc6Framework45.FunctionalTests/project.json
```diff
@@ -1,50 +1,62 @@
 {
   "webroot": "wwwroot",
   "userSecretsId": "aspnet5-Mvc6Framework45.FunctionalTests-42f45116-363a-40d3-9c44-bc58885d6e96",
-  "version": "1.0.0-rc1-*",
+  "version": "1.0.0-rc2",
+  "compilationOptions": {
+    "preserveCompilationContext": true
+  },
   "dependencies": {
-    "EntityFramework.MicrosoftSqlServer": "7.0.0-rc1-final",
```
kartang: @davidfowl I have removed netstandard library from the dependencies in AI.AspnetCore. We still see "TypeLoadExceptions"/"FileNotFoundExceptions" when we add ApplicationInsights Nuget (the one we generate from above) to a new MVC application (created using Visual studio CLI tools, with DnxCore50 framework). Please let me know if we are missing anything with respect to that setup. Thanks!

NuGet.config
```diff
@@ -3,7 +3,12 @@
   <packageSources>
     <add key="ASP.NET 5 Dev" value="https://www.myget.org/F/aspnetvnext/" />
     <add key="ASP.NET 5" value="https://www.myget.org/F/aspnetrelease/" />
-    <add key="nuget.org feed" value="https://www.nuget.org/api/v2/" />    
+    <add key="dotnet.myget.org dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/" />
+    <add key="AI myget" value="https://www.myget.org/F/applicationinsights/api/v3/index.json" />
+    <add key="dotnet.myget.org dotnet-cli" value="https://dotnet.myget.org/F/dotnet-cli/" />
+    <add key="coreclr-xunit" value="https://www.myget.org/F/coreclr-xunit/api/v2" />
+    <add key="AspNetCIDev" value="https://www.myget.org/F/aspnetcidev/api/v3/index.json" />
```
Eilon: This nuget.config file has several ASP.NET MyGet feeds on it - that can't be right. You can't mix 'dev' feeds and 'release' feeds. And for that matter, it's mixing different "qualities" of feeds: the "aspnetcidev" feed is untested 'dev' builds (RTM), and the "aspnetrelease" feed is tested and signed 'release' builds for RC2.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
+﻿{
+  "title": "Application Insights for ASP.NET Core Web Applications",
+  "description": "Application Insights for ASP.NET Core web applications. See  https://azure.microsoft.com/en-us/documentation/articles/app-insights-asp-net-five/ for more information.",
```
Eilon: Nit: There are two spaces after the word "See" (before the URL).

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
...
+  "authors": [ "Microsoft" ],
+  "version": "1.0.0-rc2-beta1",
```
Eilon: That's an interesting version number... is it RC2 or beta1?

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -0,0 +1,50 @@
...
+  "copyright": "Copyright © Microsoft. All Rights Reserved.",
+  "projectUrl": "https://github.com/Microsoft/ApplicationInsights-aspnetcore",
+  "licenseUrl": "http://go.microsoft.com/fwlink/?LinkID=510709",
+  "requireLicenseAcceptance": true,
+  "iconUrl": "http://appanacdn.blob.core.windows.net/cdn/icons/aic.png",
+  "tags": [ "Analytics", "ApplicationInsights", "Telemetry", "AppInsights", "ASP.NET Core" ],
```
Eilon: ASP.NET Core is using the tag 'aspnetcore'.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
kartang: Its actually RC2. I just added beta1 temporarily.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
kartang: Will fix it.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -40,6 +43,9 @@ public static IApplicationBuilder UseApplicationInsightsExceptionTelemetry(this
 
         public static void AddApplicationInsightsTelemetry(this IServiceCollection services, IConfiguration config)
         {
+            services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
+            services.AddSingleton<IActionContextAccessor, ActionContextAccessor>();
```
Eilon: Why does AppInsights need to add these? Doesn't ASP.NET provide these?

NuGet.config
kartang: That makes sense. Removed everything except aspnetrelease!

test/EmptyApp.FunctionalTests/project.json
```diff
@@ -1,31 +1,37 @@
...
-    "web": "Microsoft.AspNet.Hosting --server Microsoft.AspNet.Server.WebListener --server.urls http://localhost:5000",
-    "test": "xunit.runner.dnx"
+    "web": "Microsoft.AspNetCore.Hosting --server Microsoft.AspNetCore.Server.WebListener --server.urls http://localhost:5000",
```
Eilon: WebListener is not going to ship at RTM. Recommend changing this to use Kestrel if you need a standalone host.

test/EmptyApp.FunctionalTests/project.json
```diff
@@ -1,31 +1,37 @@
...
+    "preserveCompilationContext": true
+  },
   "dependencies": {
-    "Microsoft.AspNet.Server.WebListener": "1.0.0-rc1-final",
-    "Microsoft.ApplicationInsights.AspNet": "1.0.0-rc1-final",
-    "FunctionalTestUtils": "1.0.0-rc1-final",
-    "xunit.runner.dnx": "2.1.0-rc1-*",
-    "xunit": "2.1.0-rc1-*",
-    "Microsoft.AspNet.Diagnostics": "1.0.0-rc1-final"
+    "Microsoft.AspNetCore.Server.WebListener": "0.1.0-rc2-*",
```
Eilon: See comment below - WebListener is not shipping in RTM.

test/EmptyApp.FunctionalTests/Startup.cs
```diff
@@ -21,7 +23,9 @@ public class Startup
         // For more information on how to configure your application, visit http://go.microsoft.com/fwlink/?LinkID=398940
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddInstance<ITelemetryChannel>(new BackTelemetryChannel());
+            services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
+            services.AddSingleton<IActionContextAccessor, ActionContextAccessor>();
```
Eilon: Apps should not need to add these services.

test/EmptyApp.FunctionalTests/Startup.cs
```diff
@@ -31,7 +35,7 @@ public void ConfigureServices(IServiceCollection services)
         public void Configure(IApplicationBuilder app)
         {
             app.UseApplicationInsightsRequestTelemetry();
-            app.UseDeveloperExceptionPage(new ErrorPageOptions());
+            app.UseDeveloperExceptionPage(new DeveloperExceptionPageOptions());
```
Eilon: Probably don't need to pass any param to this at all if you just want the default settings.

test/FunctionalTestUtils/InProcessServer.cs
```diff
@@ -53,14 +52,15 @@ private BackTelemetryChannel Start(string assemblyName)
...
+                .UseServer("Microsoft.AspNetCore.Server.WebListener")
```
Eilon: See other comments; avoid using WebListener.

test/FunctionalTestUtils/project.json
```diff
@@ -1,31 +1,43 @@
...
-    "Microsoft.ApplicationInsights": "2.0.0-*",
-    "Microsoft.ApplicationInsights.AspNet": "1.0.0-rc1-final",
-    "Microsoft.AspNet.Http.Abstractions": "1.0.0-rc1-final",
-    "Microsoft.AspNet.Server.WebListener": "1.0.0-rc1-final",
-    "Microsoft.Dnx.Runtime": "1.0.0-rc1-final",
-    "xunit.runner.dnx": "2.1.0-rc1-*",
-    "xunit": "2.1.0-rc-*",
-    "System.Net.Http": "4.0.0-beta-*"
+    "Microsoft.AspNetCore.Http.Abstractions": "1.0.0-rc2-*",
+    "Microsoft.AspNetCore.Server.WebListener": "0.1.0-rc2-*",
```
Eilon: Remove WebListener.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
Eilon: Ok apparently ASP.NET doesn't add these, so ignore my comment.

---
In reply to: [59233587](https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/142#discussion_r59233587) [](ancestors = 59233587)

test/EmptyApp.FunctionalTests/Startup.cs
Eilon: Ok apparently ASP.NET doesn't add these, so ignore my comment.

---
In reply to: [59233935](https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/142#discussion_r59233935) [](ancestors = 59233935)

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.xproj
```diff
@@ -4,21 +4,26 @@
     <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
   </PropertyGroup>
+  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''" />
```
kartang: @Eilon Please let me know if this change is required. I am also having some trouble building the solution from visual studio. Thanks!

test/Mvc6Framework45.FunctionalTests/Startup.cs
```diff
@@ -124,8 +123,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
             {
                 routes.MapRoute(
                     name: "default",
-                    template: "{controller}/{action}/{id?}",
-                    defaults: new { controller = "Home", action = "Index" });
+                    template: "{controller=Home}/{actionIndex}/{id?}");
```
Eilon: Should be `action=Index`?

test/Mvc6Framework45.FunctionalTests/Models/AccountViewModels.cs
```diff
@@ -4,6 +4,7 @@
 using System.Threading.Tasks;
 using System.ComponentModel.DataAnnotations;
 using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNetCore.Mvc.Rendering;
```
Eilon: Hmm how can both the old namespace and new namespace work? Does this project compile?

kartang: That's weird. I didn't notice that. The project however compiles fine

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.xproj
Eilon: Hmm I don't know. Bill Hiebert and Barry Tang would be more likely to know.

---
In reply to: [59234797](https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/142#discussion_r59234797) [](ancestors = 59234797)

test/Mvc6Framework45.FunctionalTests/Models/AccountViewModels.cs
Eilon: If it compiles fine, that worries me *a lot*. That means there's a problem in the build system.

---
In reply to: [59235688](https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/142#discussion_r59235688) [](ancestors = 59235688)

test/Mvc6Framework45.FunctionalTests/project.json
```diff
@@ -1,50 +1,62 @@
...
-    "Microsoft.AspNet.StaticFiles": "1.0.0-rc1-final",
-    "Microsoft.AspNet.Tooling.Razor": "1.0.0-rc1-final",
-    "Microsoft.Extensions.Configuration.Json": "1.0.0-rc1-final",
-    "Microsoft.Extensions.Configuration.UserSecrets": "1.0.0-rc1-final",
-    "Microsoft.Extensions.CodeGenerators.Mvc": "1.0.0-rc1-final",
-    "Microsoft.Extensions.Logging": "1.0.0-rc1-final",
-    "Microsoft.Extensions.Logging.Console": "1.0.0-rc1-final",
-    "xunit.runner.dnx": "2.1.0-rc1-*",
-    "xunit": "2.1.0-rc1-*"
+    "EntityFramework.MicrosoftSqlServer": "7.0.0-rc2-*",
```
Eilon: Change to Microsoft.EntityFrameworkCore.*, version 1.0.0-rc2-*.

test/Mvc6Framework45.FunctionalTests/Models/AccountViewModels.cs
kartang: I will address the comments and see if it compiles fine with both being there. If that's the case, I will open a bug. Thanks!

NuGet.config
dnduffy: When it is released won't it be on the regular nuget.org?  I don't see that in the list anymore.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
dnduffy: We should add it.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/WebSessionTelemetryInitializerTests.cs
```diff
@@ -1,20 +1,20 @@
-﻿namespace Microsoft.ApplicationInsights.AspNet.Tests.TelemetryInitializers
+﻿namespace Microsoft.ApplicationInsights.AspNetCore.Tests.TelemetryInitializers
 {
     using System;
-
-    using Microsoft.ApplicationInsights.AspNet.TelemetryInitializers;
-    using Microsoft.ApplicationInsights.AspNet.Tests.Helpers;
+    using System.Threading.Tasks;
```
dnduffy: In another AI project I noticed #ifdefs around this using, are we going to run into that with the framework versions supported in this library?

kartang: I am not sure if I understood that right. We need to test initializer in both the frameworks. There was never an ifdefs around this test. Please let me know if I got that wrong.

NuGet.config
kartang: We will apparently add nuget feed when we publish in nuget, and remove all the local feeds.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/WebSessionTelemetryInitializerTests.cs
dnduffy: Just the "using System.Threading.Tasks;" not the tests in general.

NuGet.config
Eilon: Everything will eventually be on NuGet.org.

test/Mvc6Framework45.FunctionalTests/Startup.cs
Eilon: BTW an error like this would lead me to believe that this test doesn't run, because there's no way this would work. MVC actions wouldn't be dispatched. Can you double check that all the tests are running as expected?

kartang: Tests are not running at all as of now. That is where we are stuck. The test applications compile fine, however the requests are not going out. We changed this recently, as we couldn't figure out the problem. Will fix this and see if this could be the reason behind the failure of test application. Thank you!

test/Mvc6Framework45.FunctionalTests/Models/AccountViewModels.cs
kartang: @Eilon I had the line there, and did "dotnet restore" and "dotnet build" and seems like the build passed with no errors.

Eilon: That means that something is bringing in both the old MVC package and the new MVC package. Check the dependency graph of packages (you can try searching the project.lock.json file) to see what's bringing in Microsoft.AspNet.Mvc 6.x.

kartang: I don't see Microsoft.AspNet.Mvc package at all in project.lock.json. Furthermore, the only aspnet dependency under .nugget/packages that I see is AspNet.WebApi.Client (used for webapi test application as opposed to the current application).

Eilon: In VS can you "type in" that namespace in a CS file, see what types are there, and then Right-Click, Go To Definition, and see what assembly those are coming from? This is very suspicious...

test/EmptyApp.FunctionalTests/project.json
```diff
@@ -1,31 +1,30 @@
 ﻿{
   "webroot": "wwwroot",
-  "version": "1.0.0-*",
+  "version": "1.0.0-rc2",
```
Eilon: BTW your own tests shouldn't need a version number. Can safely remove this from all your test projects unless they actually need to get published to a NuGet feed somewhere (which is generally unlikely).

kartang: That makes sense, and yes we don't publish it anywhere. I will remove that. Thanks!

kartang: All the tests except MVC  test applications are passing now. Seems like, I am missing some configuration change/update with respect to MVC test application.

test/Mvc6Framework45.FunctionalTests/Models/AccountViewModels.cs
kartang: I am unable to navigate to symbols. Getting: Cannot navigate to the symbol under the caret.

Eilon: Hmm ok, does that seem weird? It seems weird to me...

kartang: yeah, I am totally lost on why and where is it picking the ASPNET package from. I will open a bug with respect to this scenario.

I was looking at some of the sample cli applications and they started using "netcoreapp1.0" TFM. Please let me know if that is the way to go, and should I change the TFM from "netstandardapp1.5" to the new one. Thanks.

Eilon: I'm not an expert on those TFMs, but my understanding is that you should pretend like `netstandardapp` never existed, and use `netcoreapp` instead, though they aren't *exactly* equivalent. This is where @davidfowl could likely help more.

kartang:
Will continue to work on this, and create separate pull requests with minor modifications/updates.

kartang:
The package seems to work fine for a new netcoreapp application. When installed, I see telemetry going out. The remaining changes will be mostly around updating the test applications to the latest TFMs and other settings!

PR #143, opened 2016-03-17
Added backwards compatibility for creating TelemetryClient without using dependency injection.

dnduffy:
Added backwards compatibility for creating TelemetryClient without using dependency injection.

### DIFF
--- a/test/EmptyApp.FunctionalTests/Startup.cs
+++ b/test/EmptyApp.FunctionalTests/Startup.cs
@@ -25,7 +25,7 @@ namespace EmptyApp.FunctionalTests
 
             var builder = new ConfigurationBuilder();
             builder.AddApplicationInsightsSettings(instrumentationKey: "Foo");
-            services.AddApplicationInsightsTelemetry(builder.Build());
+            services.AddApplicationInsightsTelemetry(builder.Build(), false);
         }
 
         public void Configure(IApplicationBuilder app)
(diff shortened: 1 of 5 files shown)


SergeyKanzhelev:
CC: @abaranch @karting

Looks good for me.

AlexBulankou:
:shipit:

test/EmptyApp.FunctionalTests/Startup.cs
```diff
@@ -25,7 +25,7 @@ public void ConfigureServices(IServiceCollection services)
 
             var builder = new ConfigurationBuilder();
             builder.AddApplicationInsightsSettings(instrumentationKey: "Foo");
-            services.AddApplicationInsightsTelemetry(builder.Build());
+            services.AddApplicationInsightsTelemetry(builder.Build(), false);
```
kartang: Can we leave this to true, and create a new functional test that would test if telemetry client is sending out the telemetry right? The telemetry client that we use today in the test is the one that we  get from services. At the least, we can send duplicate telemetry from two telemetry clients, and verify if we get it right.

kartang:
Looks good!

test/EmptyApp.FunctionalTests/Startup.cs
dnduffy: We'll create a separate pull request for additional functional tests.

Issue #29, opened 2015-04-16
Read version information from project.json


rinatkar:
Implement telemetry initializer that populates ```telemetry.Component.Version``` to the value read from project.json. 
This was blocked before by the issue in Configuration package. More details: https://github.com/aspnet/Configuration/issues/115

gzepeda:
One question, should the project.json file be loaded as a Configuration File on ASP.NET Core 1 using a ConfigurationBuilder or should it be loaded in a different way?

Trying to figure out the best way to load the file and get the config information from it.

SergeyKanzhelev:
Are there any other options how to load it? Will `ConfigurationBuilder` require a new dependency for the package?

SergeyKanzhelev:
I don't see problems with `ConfigurationBuilder`

kartang:
I don't think we need to explicitly load project.json. We need to make sure that "appsettings.json" is loaded, which contains the instrumentation key.

gzepeda:
I will try ConfigurationBuilder then. Just need to add an additional nuget to the project: Microsoft.Extensions.Configuration.Json. 

Does the appsettings.json contain version info?

kartang:
I am sorry for the confusion. I got the question wrong. No, appsettings will not contain the version info.

xperiandri:
Does this work without any configuration? Or I need to set up this to work?

SergeyKanzhelev:
It should work out of the box. Do you see otherwise?

Here are the pointers in code if it would help: [reading](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/9287d7e9e900f6a398c5f2b65fb2982302fd48d1/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs#L263) and [setting](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/ce6b4caade8566a451b3c98a10bc077d021f494f/src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/ComponentVersionTelemetryInitializer.cs#L11)

PR #144, opened 2016-03-21
Commit for #29 - Read version information from project.json

gzepeda:
Here are the changes for issue #29 

There is a small change to the project.json file, I needed a dependency for Microsoft.Extensions.Configuration.Json and also change the version of Microsoft.ApplicationInsights so the project would compile.


kartang:
Looks good to me!

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/ComponentVersionTelemetryInitializer.cs
```diff
@@ -0,0 +1,38 @@
...
+    using Microsoft.Extensions.Configuration;
+
+    /// <summary>
+    /// A telemetry initializer that populates telemetry.Context.Component.Version to the value read from project.json
+    /// </summary>
+    public class ComponentVersionTelemetryInitializer : TelemetryInitializerBase
+    {
+        private const string versionConfigurationOption = "dependencies:Microsoft.ApplicationInsights.AspNet";
+
+        public ComponentVersionTelemetryInitializer(IHttpContextAccessor httpContextAccessor):base(httpContextAccessor)
```
SergeyKanzhelev: Can you get IConfiguration object using dependency injection here and than try to do: 
```
config[versionConfigurationOption].ToString();
```
instead of guessing which file contains configuration?

gzepeda: Sure. I will try that approach.

gzepeda: I have made the change, but before creating a new Pull request. I wanted to ask if the approach was correct. 

I added this to the AddApplicationInsightsTelemetry method:

```
var projectJsonConfiguration = new ConfigurationBuilder().AddJsonFile("project.json").Build();
services.AddInstance<IConfiguration>(projectJsonConfiguration);
```

And Injecting the IConfiguration dependency on the constructor for the created class. 

It does not conflict with the TelemetryConfiguration but I don't feel it is the right place to have that service declared.

gzepeda:
I have made the change, but before creating a new Pull request. I wanted to ask if the approach was correct. 

I added this to the AddApplicationInsightsTelemetry method:

```
var projectJsonConfiguration = new ConfigurationBuilder().AddJsonFile("project.json").Build();
services.AddInstance<IConfiguration>(projectJsonConfiguration);
```

And Injecting the IConfiguration dependency on the constructor for the created class. 

It does not conflict with the TelemetryConfiguration but I don't feel it is the right place to have that service declared.

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -48,6 +49,9 @@ public static void AddApplicationInsightsTelemetry(this IServiceCollection servi
             services.AddSingleton<ITelemetryInitializer, WebSessionTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebUserTelemetryInitializer>();
 
+            var projectJsonConfiguration = new ConfigurationBuilder().AddJsonFile("project.json").Build();
```
SergeyKanzhelev: Configuration is already passed as an argument. You can pass it into initializer constructor

gzepeda:
Done. 

One more thing: I am not sure how to reuse the service on the unit tests without declaring a new instance and sending it to the initializer.

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -48,6 +49,8 @@ public static void AddApplicationInsightsTelemetry(this IServiceCollection servi
             services.AddSingleton<ITelemetryInitializer, WebSessionTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebUserTelemetryInitializer>();
 
+            services.AddInstance<IConfiguration>(config);
```
SergeyKanzhelev: You can just pass this configuration to `ComponentVersionTelemetryInitializer` constructor on the line 44

src/Microsoft.ApplicationInsights.AspNet/project.json
```diff
@@ -14,14 +14,15 @@
...
+        "Microsoft.ApplicationInsights": "2.0.0-beta4",
         "Microsoft.AspNet.Hosting.Abstractions": "1.0.0-rc1-final",
         "Microsoft.AspNet.Http.Abstractions": "1.0.0-rc1-final",
         "Microsoft.AspNet.Http.Features": "1.0.0-rc1-final",
         "Microsoft.AspNet.Mvc": "6.0.0-rc1-final",
         "Microsoft.Extensions.Logging.Abstractions": "1.0.0-rc1-final",
         "Microsoft.Extensions.Configuration": "1.0.0-rc1-final",
-        "Microsoft.Extensions.DiagnosticAdapter": "1.0.0-rc1-final"
+        "Microsoft.Extensions.DiagnosticAdapter": "1.0.0-rc1-final",
+        "Microsoft.Extensions.Configuration.Json": "1.0.0-rc1-final"
```
SergeyKanzhelev: we don't need this new reference with this implementation

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/ComponentVersionTelemetryInitializer.cs
```diff
@@ -0,0 +1,40 @@
...
+    using DataContracts;
+    using Microsoft.AspNet.Http;
+    using Microsoft.Extensions.Configuration;
+
+    /// <summary>
+    /// A telemetry initializer that populates telemetry.Context.Component.Version to the value read from project.json
+    /// </summary>
+    public class ComponentVersionTelemetryInitializer : TelemetryInitializerBase
+    {
+        private const string _versionConfigurationOption = "dependencies:Microsoft.ApplicationInsights.AspNet";
```
SergeyKanzhelev: don't we need just 'version' ?

SergeyKanzhelev: I think we need this: https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/master/test/WebApiShimFw46.FunctionalTests/project.json#L3

src/Microsoft.ApplicationInsights.AspNet/project.json
gzepeda: Removed.

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/ComponentVersionTelemetryInitializer.cs
gzepeda: There is no "version" entry on the project.json file of the test project

\test\Microsoft.ApplicationInsights.AspNet.Tests\project.json

I can assume there will always be one or return null.

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/ComponentVersionTelemetryInitializer.cs
```diff
@@ -0,0 +1,40 @@
...
+                _configuration = configuration;
+            }
+        }
+
+        protected override void OnInitializeTelemetry(HttpContext platformContext, RequestTelemetry requestTelemetry, ITelemetry telemetry)
+        {
+            if (string.IsNullOrEmpty(telemetry.Context.Component.Version))
+            {
+                if (_configuration == null) {                
+                    telemetry.Context.Component.Version = null;
```
SergeyKanzhelev: Just don't set it. No need to set to null

src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/ComponentVersionTelemetryInitializer.cs
SergeyKanzhelev: If there is no version in specific project we can return null. It means there is no version that we are sure about and customer will need to implement the logic of setting version himself

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
gzepeda: Should I do something like this:
```
            services.AddSingleton<ITelemetryInitializer>(ComponentVersionTelemetryInitializer =>
            {
                return new ComponentVersionTelemetryInitializer(null, config);
            });
```
The problem I see is not having the httpContextAccessor

SergeyKanzhelev: Yes, please

gzepeda: A bunch of Unit Tests produce exceptions if I send the first parameter (httpContextAccessor as null). Even if I don't need it on the initializer, as long as the class implements Interface TelemetryInitializerBase it is kind of mandatory to send it not null. 

```
\test\Microsoft.ApplicationInsights.AspNet.Tests\Extensions\Appli
cationInsightsExtensionsTests.cs(52,0): at Microsoft.Extensions.DependencyInjection.ApplicationInsightsExtensionsTests.AddApplicationI
nsightsTelemetry.RegistersExpectedServices(Type serviceType, Type implementationType, ServiceLifetime lifecycle)
      System.ArgumentNullException : Value cannot be null.
      Parameter name: httpContextAccessor
```

```
\test\Microsoft.ApplicationInsights.AspNet.Tests\Extensions\Appli
cationInsightsExtensionsTests.cs(275,0): at Microsoft.Extensions.DependencyInjection.ApplicationInsightsExtensionsTests.AddApplication
InsightsSettings.RegistersTelemetryConfigurationFactoryMethodThatReadsInstrumentationKeyFromSettings()
      System.ArgumentNullException : Value cannot be null.
      Parameter name: httpContextAccessor
```

SergeyKanzhelev: Just inherit it from `ITelemetryInitializer`, not `TelemetryInitializerBase`

gzepeda: Done. Let me know if the latest commits work.

src/Microsoft.ApplicationInsights.AspNet/project.json
```diff
@@ -14,7 +14,7 @@
         "delaySign": true
     },
     "dependencies": {
-        "Microsoft.ApplicationInsights": "2.0.0-*",
+        "Microsoft.ApplicationInsights": "2.0.0-beta4",
```
kartang: Why are we specifically referring to beta4 here?

dnduffy: Thank-you for all of your changes, unless there is a problem with 2.0.0-* please go ahead and change it and then we will complete the pull request.

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
dnduffy: Unit tests all pass.

src/Microsoft.ApplicationInsights.AspNet/project.json
gzepeda: Done.

PR #145, opened 2016-03-23
Fix request name and add framework setup to sdkversion

kartang:


src/Microsoft.ApplicationInsights.AspNet/TelemetryInitializers/TelemetryInitializerBase.cs
```diff
@@ -1,16 +1,15 @@
 ﻿namespace Microsoft.ApplicationInsights.AspNet.TelemetryInitializers
 {
-    using System;
-    using System.Diagnostics;
```
kartang: Will fix these gaps. I believe I should change the default tab size.

src/Microsoft.ApplicationInsights.AspNet/SdkVersionUtils.cs
```diff
@@ -5,11 +5,27 @@
...
+        internal const string VersionPrefix = "aspnet5-";
+
+        internal static string GetFrameworkType()
+        {
+            string framework;
+#if dnx451
+           // F representing the full framework
+           framework = "F:"; 
+#elif dnxcore50
+          // C representing the core framework
```
dnduffy: CTRL-K CTRL-D in VS to make indenting consistent.

dnduffy:
:+1:

abaranch:
:shipit:

src/Microsoft.ApplicationInsights.AspNet/SdkVersionUtils.cs
```diff
@@ -1,17 +1,33 @@
...
+#if dnx451
+           // F representing the full framework
+           framework = "F:"; 
+#elif dnxcore50
+            // C representing the core framework
+            framework = "C:";
+#else
+          // O representing other framework
+          framework = "O:";
+#endif
```
AlexBulankou: What can other be?

kartang: The reason I kept there for now, is because we might have netstandard1.3, which can support "dotnet5.4" application for example, or "netstandardapp1.5". I am not sure how is it going to take into account all those versions. We can remove that once we are sure about this details.

PR #152, opened 2016-04-07
Changes to MVC template to match the latest updates

kartang:


test/Mvc6Framework45.FunctionalTests/bower.json
```diff
@@ -1,12 +0,0 @@
-{
-  "name": "ASP.NET",
-  "private": true,
-  "dependencies": {
-    "bootstrap": "3.0.0",
```
dnduffy: Have any local copies of these dependencies been removed as well?

dnduffy:
:shipit:

test/Mvc6Framework45.FunctionalTests/bower.json
kartang: we will not store any local copies of dependencies

PR #157, opened 2016-04-18
Adding server telemetry channel, adaptive sampling and making Active as default configuraiton.

kartang:
Adding server telemetry channel, enabling sampling (with default adaptive) and removing existing configure option to create config.

- TelemetryConfiguration.Active is used if services do not have an existing configuration instance, in which case it picks the configuration from the services and updates it.
- Server telemetry channel is assigned to `config.TelemetryChannel` only in full framework, and if no other telemetry channel exists in services (in which case, `config.TelemetryChannel` will be the existing channel in services).
- Neither the channel nor the sampling processor is added to the services collection. 
- Customers have access to telemetry configuration, and can add his own custom telemetry processors through chain builder. 
- Customers can disable sampling using the parameter `DisableSampling` when adding Application Insights through services -> `services.AddApplicationInsightsTelemetry(config, true)`. This will not add `AdaptiveSamplingTelemetryProcessor` which is added by default in the other case.
- Existing boolean parameter on `services.AddApplicationInsightsTelemetry` to control the usage of `TelemetryConfiguration.Active` vs `TelemetryConfiguration.CreateDefault()` is removed and `Active` is used in all the scenarios.

**Note: By default, now, server telemetry channel will be used to send the telemetry, as opposed to InMemoryChannel. If customers have to use their specific channel, they can add channel to the services, and we pick that instead of our default channels.**

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,8 +1,8 @@
 ﻿{
     "sources": [ "src", "test" ],
     "sdk": {
-            "architecture": "x86",
+      "architecture": "x86",
       "runtime": "clr",
-      "version": "1.0.0-rc1-update1"
+      "version": "1.0.0-rc1-update2"
     }
 }
--- a/src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
@@ -40,10 +40,10 @@
             return app.UseMiddleware<ExceptionTrackingMiddleware>();
         }
 
-        public static void AddApplicationInsightsTelemetry(this IServiceCollection services, IConfiguration config)
+        public static void AddApplicationInsightsTelemetry(this IServiceCollection services, IConfiguration config, bool DisableSampling = false)
         {
             services.AddSingleton<ITelemetryInitializer, DomainNameRoleInstanceTelemetryInitializer>();
-            services.AddSingleton<ITelemetryInitializer>(ComponentVersionTelemetryInitializer =>
+            services.AddSingleton<ITelemetryInitializer>(serviceProvider =>
             {
                 return new ComponentVersionTelemetryInitializer(config);
             });
@@ -63,8 +63,8 @@
 
             services.AddSingleton<TelemetryConfiguration>(serviceProvider =>
             {
-                var telemetryConfiguration = serviceProvider.GetService<TelemetryConfiguration>() ?? TelemetryConfiguration.Active;
-                AddServerTelemetryChannelAndSamplingForFullFramework(serviceProvider, telemetryConfiguration);
+                var telemetryConfiguration = TelemetryConfiguration.Active;
+                AddServerTelemetryChannelAndSamplingForFullFramework(serviceProvider, telemetryConfiguration, DisableSampling);
                 telemetryConfiguration.TelemetryChannel = serviceProvider.GetService<ITelemetryChannel>() ?? telemetryConfiguration.TelemetryChannel;
                 AddTelemetryConfiguration(config, telemetryConfiguration);
                 AddServicesToCollection(serviceProvider, telemetryConfiguration.TelemetryInitializers);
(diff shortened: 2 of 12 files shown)


global.json
```diff
@@ -1,8 +1,8 @@
 ﻿{
     "sources": [ "src", "test" ],
     "sdk": {
-            "architecture": "x86",
+      "architecture": "x86",
       "runtime": "clr",
-      "version": "1.0.0-rc1-update1"
+      "version": "1.0.0-rc1-update2"
```
dnduffy: Are we were up to update3 in some branch?

kartang: This version corresponds to the dnx runtime. They have updated the run time version, and I referred to the latest available one.

test/Microsoft.ApplicationInsights.AspNet.Tests/Extensions/ApplicationInsightsExtensionsTests.cs
```diff
@@ -302,6 +253,51 @@ public static void RegistersTelemetryConfigurationFactoryMethodThatPopulatesItWi
...
+                var services = ApplicationInsightsExtensionsTests.GetServiceCollectionWithContextAccessor();
+                var config = new ConfigurationBuilder().AddApplicationInsightsSettings(endpointAddress: "http://localhost:1234/v2/track/").Build();
+                services.AddApplicationInsightsTelemetry(config, true);
+                IServiceProvider serviceProvider = services.BuildServiceProvider();
+                var telemetryConfiguration = serviceProvider.GetRequiredService<TelemetryConfiguration>();
+                Assert.Equal(telemetryConfiguration.TelemetryProcessors.Count, exisitingProcessorCount);
+            }
+
+            [Fact]
+            public static void AddsServerTelemetryChannelInFullFramework()
```
dnduffy: Doesn't this require the #ifdef for full framework around the test?

kartang: We have it, grouped for all the tests in full framework. A combined ifdef.

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -181,5 +184,22 @@ private static void InitializeModulesWithConfiguration(IServiceProvider serviceP
                 service.Initialize(configuration);
             }
         }
+
+        private static void AddServerTelemetryChannelAndSamplingForFullFramework(IServiceProvider serviceProvider, TelemetryConfiguration configuration, bool DisableSampling)
```
dnduffy: I'm wondering if instead of having a flag and making an internal call for using adaptive sampling if we should just have the users make a call or pass the class they want to use as a parameter.

kartang: As per the offline discussion,we will leave this as is for now, and when there is a change in Core/channels to remove processors, we can completely ignore this boolean then.

dnduffy:
:shipit:

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -38,10 +40,10 @@ public static IApplicationBuilder UseApplicationInsightsExceptionTelemetry(this
             return app.UseMiddleware<ExceptionTrackingMiddleware>();
         }
 
-        public static void AddApplicationInsightsTelemetry(this IServiceCollection services, IConfiguration config, bool reuseActiveConfig = true)
+        public static void AddApplicationInsightsTelemetry(this IServiceCollection services, IConfiguration config, bool disableDefaultAdaptiveSampling = false)
```
AlexBulankou: Chatted with David, let's discuss this some more tomorrow before we merge to master.
1. Consider wrapping this flag in a config structure to allow future changes
2. Customers who onboarded to previous version with reuseActiveConfig will not be able to notice anything changed and the fact that now this flag has completely different meaning.

kartang: Sure, let us discuss about using the config. However, "reuseActiveConfig" was never available for customers. It is always in develop, and we planned to have it in this merge to master.

PR #160, opened 2016-04-20
Tests fail in coreclr framework (in build), as TelemetryCOnfiguration…

kartang:
Tests fail in coreclr framework (in build), as TelemetryConfiguratin.Active is messed with fake telemetry channels. If the environment is coreclr, we just use the existing telemetry channel, thereby causing the tests to fail in the build. Adding a channel to services, will resolve as, AI picks the channel from services, if any such exists.


dnduffy:
:shipit:

PR #161, opened 2016-04-21
Daduffy/add event source logging

dnduffy:


src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLogger.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// Delegate of basic event source function calls.
+        /// </summary>
+        /// <param name="message">The message to log.</param>
+        /// <param name="appDomainName">An ignored placeholder to make EventSource happy.</param>
+        private delegate void EventSourceLogCall(string message, string appDomainName = "Incorrect");
+
+        /// <summary>
+        /// Gets an instance of the telemetry logger.
+        /// </summary>
+        public static TelemetryLogger Instance
```
kartang: nit-pick: "Instance" seems to be used in event source, and in this code as well.

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -104,6 +110,12 @@ public static IConfigurationBuilder AddApplicationInsightsSettings(this IConfigu
                 wasAnythingSet = true;
             }
 
+            if (level != null)
+            {
+                TelemetryLogger.Instance.Level = level.Value;
+                wasAnythingSet = true;
```
kartang: we are not adding anything to telemetryConfigurationSource. Please let me know if we still need to set wasAnythingSet to be true?

kartang:
:shipit:

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLogger.cs
```diff
@@ -0,0 +1,120 @@
...
+    using System.Diagnostics.Tracing;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// Application Insights internal telemetry logging implementation.
+    /// </summary>
+    internal sealed class TelemetryLogger : ILogger
+    {
+        /// <summary>Instance of </summary>
+        private static TelemetryLogger instance = new TelemetryLogger();
```
AlexBulankou: "private static readonly" to semantically enforce singleton?

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLogger.cs
```diff
@@ -0,0 +1,120 @@
...
+            };
+        }
+
+        /// <summary>
+        /// Delegate of basic event source function calls.
+        /// </summary>
+        /// <param name="message">The message to log.</param>
+        /// <param name="appDomainName">An ignored placeholder to make EventSource happy.</param>
+        private delegate void EventSourceLogCall(string message, string appDomainName = "Incorrect");
+
```
AlexBulankou: Why do we need this delegate? Looks like you're populating an array of these objects, (like eventSource.LogAlways), can we use their type?

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLogger.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        /// <summary>
+        /// Checks if the given log level is enabled.
+        /// </summary>
+        /// <param name="logLevel">The leg level to check.</param>
+        /// <returns>True if the specified log level is enabled, otherwise false.</returns>
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return logLevel >= this.Level && this.Level != LogLevel.None && logLevel != LogLevel.None;
```
AlexBulankou: I think if we have the first check, then one of 2 subsequents checks is not needed

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLoggerScope.cs
```diff
@@ -0,0 +1,37 @@
...
+//-----------------------------------------------------------------------
+namespace Microsoft.ApplicationInsights.AspNet.Extensibility.Implementation.Tracing
+{
+    using System;
+
+    /// <summary>
+    /// Logging scope for <code>TelemetryLogger.BeginScopeImpl</code> calls.
+    /// </summary>
+    public class TelemetryLoggerScope : IDisposable
+    {
```
AlexBulankou: Should you follow basic dispose pattern (https://msdn.microsoft.com/en-us/library/b1yfkh5e(v=vs.110).aspx)

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLoggerScope.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <summary>State of the logging scope instance.</summary>
+        private bool isDisposed = false;
+
+        /// <summary>
+        /// Gets a value indicating whether or not the logging scope instance is disposed.
+        /// </summary>
+        internal bool IsDisposed
```
AlexBulankou: Why do we need to expose disposed state?

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLoggerScope.cs
```diff
@@ -0,0 +1,37 @@
...
+// </copyright>
+//-----------------------------------------------------------------------
+namespace Microsoft.ApplicationInsights.AspNet.Extensibility.Implementation.Tracing
+{
+    using System;
+
+    /// <summary>
+    /// Logging scope for <code>TelemetryLogger.BeginScopeImpl</code> calls.
+    /// </summary>
+    public class TelemetryLoggerScope : IDisposable
```
AlexBulankou: Perhaps add some unit test to for TelemetryLoggerScope?

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/AspNetEventSource.cs
```diff
@@ -0,0 +1,121 @@
...
+namespace Microsoft.ApplicationInsights.AspNet.Extensibility.Implementation.Tracing
+{
+    using System;
+    using System.Diagnostics.Tracing;
+
+    /// <summary>
+    /// Event source for Application Insights ASP.NET 5 SDK.
+    /// </summary>
+    [EventSource(Name = "Microsoft-ApplicationInsights-AspNet")]
+    internal sealed class AspNetEventSource : EventSource
```
AlexBulankou: Similarly how we are doing for [Core](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/1ccd5b466451db1884b242b6a58b3e9401a1fbcf/src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs), I think we want to differentiate each diagnostic message source with a separate event ID. This way we are ensuring message reuse and enable efficient filtering and search for diagnostic events.

dnduffy: There are two very different paradigms between ILogger and EventSource.  I can make it work the way you describe but it would not be following the ILogger model and there are several reasons why I would not recommend it we can discuss.

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLogger.cs
dnduffy: I'm using it as function pointers for the C# version of a jump table.

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLogger.cs
dnduffy: Yes, "Instance" is a common name for a Singleton instance of a class.

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLogger.cs
dnduffy: The first check alone is not sufficient because "None" is defined as int.MaxValue instead of as 0 so it does not follow the sequence of the other values from Debug = 1 to Critical = 6.

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/TelemetryLoggerScope.cs
dnduffy: We should discuss what we should do with this class since it isn't really even being used, there is a bigger ILogger design question here.

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsExtensions.cs
dnduffy: Good point, removed.

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/AspNetEventSource.cs
```diff
@@ -0,0 +1,121 @@
...
+//-----------------------------------------------------------------------
+namespace Microsoft.ApplicationInsights.AspNet.Extensibility.Implementation.Tracing
+{
+    using System;
+    using System.Diagnostics.Tracing;
+
+    /// <summary>
+    /// Event source for Application Insights ASP.NET 5 SDK.
+    /// </summary>
+    [EventSource(Name = "Microsoft-ApplicationInsights-AspNet")]
```
abaranch: Should we call it AspNetCore? Without Core and without 5 that is quite generic...

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/AspNetEventSource.cs
```diff
@@ -0,0 +1,121 @@
...
+            this.WriteEvent(6, message ?? string.Empty, this.ApplicationName);
+        }
+
+        /// <summary>
+        /// Logs an event for the critical message level.
+        /// </summary>
+        /// <param name="message">The message to write an event for.</param>
+        /// <param name="appDomainName">An ignored placeholder to make EventSource happy.</param>
+        [Event((int)EventLevel.Critical, Message = "{0}", Level = EventLevel.Critical, Keywords = Keywords.Diagnostics)]
+        public void LogCritical(string message, string appDomainName = "Incorrect")
```
abaranch: Should we/Can we also have LogCritical and LogError with Keywork UserActionable? Usually configuration errors or something that customer can fix on his own are logged with this keyword. Also it UI they would have different prefixes and can be filtered.

src/Microsoft.ApplicationInsights.AspNet/project.json
```diff
@@ -33,15 +33,17 @@
           "Microsoft.ApplicationInsights.PerfCounterCollector": "2.0.0",
           "Microsoft.ApplicationInsights.DependencyCollector": "2.0.0",
           "Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel": "2.0.0"
```
abaranch: "," in the end is missing

dnduffy: Oops, I missed that when having to manually resolve a merge conflict.  Fixed.

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/AspNetEventSource.cs
dnduffy: Added "Core".

dnduffy:
I've deleted the ILogger to EventSource bridge and related code for now and converted to a pure EventSource implementation.

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/AspNetEventSource.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// </summary>
+        public string ApplicationName { [NonEvent] get; [NonEvent]private set; }
+
+        /// <summary>
+        /// Logs an event for the an exception in the TelemetryInitializerBase Initialize method.
+        /// </summary>
+        /// <param name="exception">The exception to write an event for.</param>
+        /// <param name="appDomainName">An ignored placeholder to make EventSource happy.</param>
+        [Event(1, Message = "{0}", Level = EventLevel.Error, Keywords = Keywords.Diagnostics)]
+        public void LogTelemetryInitializerBaseInitializeException(Exception exception, string appDomainName = "Incorrect")
```
abaranch: This override is very slow (https://msdn.microsoft.com/en-us/library/hh393360(v=vs.110).aspx). Do you think we can pass exception.Message, exception.StackTrace as 2 parameters here?

abaranch:
Add one comment to consider. But :shipit:

src/Microsoft.ApplicationInsights.AspNet/Extensibility/Implementation/Tracing/AspNetEventSource.cs
dnduffy: I've changed it to string so that I can reuse the EventSourceTest class from core which doesn't support non-value types.

PR #162, opened 2016-04-21
Adding Application Insights Service Options services extension

kartang:
Application Insights Service Options is an optional parameter to Application Insights extension to services. This controls the overriding behavior of the default set of features for AI.


AlexBulankou:
:shipit:

src/Microsoft.ApplicationInsights.AspNet/Extensions/ApplicationInsightsServiceOptionsExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// Extensions for Application Insights Service Options that returns specific set of properties if serviceOptions is enabled. If not, it returns the default initial value.
+    /// </summary>
+    internal static class ApplicationInsightsServiceOptionsExtensions
+    {
+        private static ApplicationInsightsServiceOptions _serviceOptions = new ApplicationInsightsServiceOptions();
+
+        /// <summary>
+        /// Checks if sampling is disabled through service options.
+        /// </summary>
+        internal static bool GetDisableDefaultSampling(this ApplicationInsightsServiceOptions serviceOptions)
```
dnduffy: Why is there an extension method for a property which is already accessible?

kartang: The reason we add this, is we can have the functionality implementation separate from the definition. For example, tomorrow we have initializers as optional features. We add the respective list in the Options. We then access these extensions to do the functionality that we intended to do. We pass the service options and services and we keep implementation abstract from options. We don't want to expose any of those details to customers. That is why, we also ignore any null checks, as the implementation details is taken care in this extension.

kartang: Today it's just a boolean, but tomorrow it can be a complex option which needs some functionality being added from our side. PLease let me know if this makes sense.

dnduffy:
:shipit:

PR #170, opened 2016-04-26
Change DisableDefaultSampling to EnableAdaptiveSampling

kartang:
EnableAdaptiveSampling is by default true and overriding it with false will not enable the sampling feature.


dnduffy:
:shipit:

PR #187, opened 2016-05-07
Enabling quickpulse metric stream by default along with the associated service option

kartang:
- Quick pulse metric stream is now enabled by default, and is the first telemetry processor. Default addition can be controlled using service option: ```EnableQuickPulseMetricStream```

- Fixing minor dependency warnings in functional tests that are causing build to fail.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
@@ -5,39 +5,25 @@
     /// </summary>
     public class ApplicationInsightsServiceOptions
     {
-        private bool enableQuickPulseMetricStream = true;
-        private bool enableAdaptiveSampling = true;
+        /// <summary>
+        /// Application Insights service options that controlls the default behavior of application insights features.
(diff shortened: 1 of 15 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
```diff
@@ -5,9 +5,26 @@
...
 
         /// <summary>
+        /// Setting EnableQuickPulseMetricStream to false, will disable the default quick pulse metric stream. As a result, QuickPulseTelemetryModule
+        /// and QuickPulseTelemetryProcessor are not registered with the configuration by default.
+        /// </summary>
+        public bool EnableQuickPulseMetricStream
+        {
+            get
+            {
+                return enableQuickPulseMetricStream;
```
dnduffy: This is good, I'm just curious why not use automatic properties?

kartang: That's a good point. Will change them to auto implementations. Thanks!

dnduffy:
:shipit:

PR #193, opened 2016-05-10
Changing the sdkversion and updating the tests

kartang:


src/Microsoft.ApplicationInsights.AspNetCore/SdkVersionUtils.cs
```diff
@@ -5,20 +5,17 @@
 
     internal class SdkVersionUtils
     {
-        internal const string VersionPrefix = "aspnetCore-";
+        internal const string VersionPrefix = "aspnet5";
```
abaranch: We can consider defining two consts under #if rather that introducing a separate method and do concatenation.

abaranch:
:shipit:

src/Microsoft.ApplicationInsights.AspNetCore/SdkVersionUtils.cs
kartang: makes sense! We can try that. I will update the code. Thanks!

test/Microsoft.ApplicationInsights.AspNetCore.Tests/SdkVersionTestUtils.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Microsoft.ApplicationInsights.AspNetCore.Tests
+{
+    public class SdkVersionTestUtils
+    {
+        public static string GetExpectedSdkVersion()
+        {
+            string expectedSdkVersion = "aspnet5";
+
+#if NET451
+        expectedSdkVersion = expectedSdkVersion + "F";
```
dnduffy: Why not call GetFrameworkType() instead of repeating the "F" and "C" logic and ifdefs?

dnduffy:
Tweak and :shipit:

test/Microsoft.ApplicationInsights.AspNetCore.Tests/SdkVersionTestUtils.cs
kartang: I thought we are testing it, and any change there would never be caught if we reuse it?

src/Microsoft.ApplicationInsights.AspNetCore/SdkVersionUtils.cs
abaranch: That would be one const with the same name. It would either have one value or the other (without runtime concatination)

PR #210, opened 2016-05-17
Fix dotnet-test-xunit dependency which causes tests to fail in net451…

kartang:
Fix dotnet-test-xunit dependency which causes tests to fail in net451 on local box.

The bits in nuget have an issue.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,6 +2,7 @@
 <configuration>
   <packageSources>
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="Aspnet Release Old" value="https://myget.org/f/aspnetrelease" />
   </packageSources>
   <disabledPackageSources />
 </configuration>
--- a/test/EmptyApp.FunctionalTests/project.json
+++ b/test/EmptyApp.FunctionalTests/project.json
@@ -10,7 +10,7 @@
     "Microsoft.AspNetCore.Diagnostics": "1.0.0-rc2-final",
     "FunctionalTestUtils": { "target": "project" },
     "Microsoft.NETCore.Platforms": "1.0.1-*",
-    "dotnet-test-xunit": "1.0.0-*",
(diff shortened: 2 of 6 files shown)


test/EmptyApp.FunctionalTests/project.json
```diff
@@ -10,7 +10,7 @@
     "Microsoft.AspNetCore.Diagnostics": "1.0.0-rc2-final",
     "FunctionalTestUtils": { "target": "project" },
     "Microsoft.NETCore.Platforms": "1.0.1-*",
-    "dotnet-test-xunit": "1.0.0-*",
+    "dotnet-test-xunit": "1.0.0-rc2-187791-49",
```
dnduffy: Out of curiosity, what is going wrong with 1.0.0-*?

NuGet.config
```diff
@@ -2,6 +2,7 @@
 <configuration>
   <packageSources>
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="Aspnet Release Old" value="https://myget.org/f/aspnetrelease" />
```
dnduffy: What do we still need from this source?

dnduffy:
:shipit:

test/EmptyApp.FunctionalTests/project.json
kartang: 1.0.0-* will bring the latest bits available. Unless we are sure that the issue is fixed in the latest version, we don't want to use it. Its a testing package, and I was told to use this package that works for sure. So, hard coded the version for this one. Furthermore, I believe this change should only stay in develop branch. Please let me know if that makes sense to have it only in develop. Thanks!

NuGet.config
kartang: nuget.org has the corrupted bit (the specified version is not available in nuget.org), which is why I have added aspnetrelease feed as well so that it brings the specific version.

PR #219, opened 2016-06-16
Updated the JavaScript snippet to the latest version…

dnduffy:
Updated the JavaScript snippet to the latest version and put it into a resource file for easier maintenance.


abaranch:
:shipit:

src/Microsoft.ApplicationInsights.AspNetCore/ApplicationInsightsJavaScript.cs
```diff
@@ -48,9 +29,7 @@ public string FullScript
             {
                 if (!string.IsNullOrEmpty(this.telemetryConfiguration.InstrumentationKey))
                 {
-                    return ScriptPrefix + @"{
-  instrumentationKey: '" + this.telemetryConfiguration.InstrumentationKey + @"'
-}" + ScriptPostfix;
+                    return string.Format(Resources.JavaScriptSnippet, this.telemetryConfiguration.InstrumentationKey);
```
SergeyKanzhelev: will it read from resources for every request or it will cache it in the string?

dnduffy: It is cached in the resource manager I believe (it's been a while since I read that code in Reflector...).

dnduffy: No, I seem to be remembering wrong, I'll cache it.

PR #229, opened 2016-06-23
Made signing check for the existence of the signing DLL.

dnduffy:
Added test projects and additional debugging messages.
Resolved xunit reference problems.


kartang:
:shipit:

PR #231, opened 2016-06-24
Changed the path references for the test projects…

dnduffy:
… and used them as the working dir for the exec dotnet test calls.


SergeyKanzhelev:
Looks ok for me. :shipit:

PR #241, opened 2016-07-07
Update JS SDK snippet

kamilszostak:


src/Microsoft.ApplicationInsights.AspNetCore/Resources.resx
```diff
@@ -119,11 +119,11 @@
   </resheader>
   <data name="JavaScriptSnippet" xml:space="preserve">
     <value>    &lt;script type="text/javascript"&gt;
-        var appInsights=window.appInsights||function(config){{
```
kamilszostak: Any reason why there were two {`s before?

dnduffy: Yes, they have to be there so that string.Format(...) doesn't think they are part of a parameter replacement like {0}.  The double {{ and }} escape the { and } just like \\ escapes a \.

kartang:
Do this PR against develop?

kamilszostak:
I'll fix the {{ }} and reopen the pull request against the develop branch. Thanks!

PR #253, opened 2016-08-25
Update JS snippet from the ApplicationInsights-JS repo

dfederm:
This updates the JS snippet from the latest in the ApplicationInsights-JS repo. I've left in the minor differences of course like the {0} for the inst key and CDN path.

See https://github.com/Microsoft/ApplicationInsights-JS/pull/295 for additional context.

PS the snippet also seems to already have been a little diverged from what was in ApplicationInsights-JS, so this change pulls those in too as a side effect. Either way it should be consistent with that repo as of now.


dnduffy:
👍

dnduffy:
I was wondering why the diff was "so different" until we realized this was against master instead of develop. ;)  I'll close this PR, please open it against develop. :)

PR #254, opened 2016-08-25
Update JS snippet from the ApplicationInsights-JS repo

dfederm:
This updates the JS snippet from the latest in the ApplicationInsights-JS repo. I've left in the minor differences of course like the {0} for the inst key and CDN path.

See Microsoft/ApplicationInsights-JS#295 for additional context.


dnduffy:
👍

PR #256, opened 2016-09-01
Adding culture and changing telemetry client to singleton

kartang:


SergeyKanzhelev:
:shipit:

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/HttpRequestExtensionsTest.cs
```diff
@@ -66,7 +67,7 @@ public void TestGetUriReturnsCorrectUriIfRequestObjectSchemeAndHostAreSpecified(
             var uri = HttpRequestExtensions.GetUri(request);
 
             Assert.Equal(
-                new Uri(string.Format("{0}://{1}", ExpectedSchema, ExpectedHostName)), 
+                new Uri(string.Format(CultureInfo.InvariantCulture, "{0}://{1}", ExpectedSchema, ExpectedHostName)),
```
dnduffy: Are you sure that this shouldn't be Ordinal rather than Invariant?

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/HttpRequestExtensionsTest.cs
```diff
@@ -99,7 +100,7 @@ public void TestGetUriReturnsCorrectUriIfRequestObjectSchemeAndHostAndPathAndQue
             var uri = HttpRequestExtensions.GetUri(request);
 
             Assert.Equal(
-                new Uri(string.Format("{0}://{1}{2}{3}", ExpectedSchema, ExpectedHostName, ExpectedPath, ExpectedQueryString)),
+                new Uri(string.Format(CultureInfo.InvariantCulture, "{0}://{1}{2}{3}", ExpectedSchema, ExpectedHostName, ExpectedPath, ExpectedQueryString)),
```
dnduffy: The schema and hostname might be safe with Invariant, although I doubt that hostname will always be safe...
However, I'm not sure that the path and query name are bound to punycode rules or anything else that will force them to be ASCII-like.  I don't recall for certain if it could be an issue with URLs or not but it warrants checking.

kartang: I thought Ordinal does a basic string comparison, and invariant culture also understands different symbols for comparison, and is more strict. I don't think there will be an adverse effect on using invariant culture here. Please correct me if I am wrong!

dnduffy: No, invariant means it can't handle anything "different", it's basically English ASCII rules.

PR #257, opened 2016-09-01
Removing unused external test dependencies and adding third party notice

kartang:


dnduffy:
Assuming all tests still pass after removing those references then :+1: .

kartang:
The tests do pass on my box!

PR #263, opened 2016-09-16
Updated .NET Core to 1.0.1.

dnduffy:


test/FunctionalTestUtils/project.json
```diff
@@ -1,18 +1,19 @@
...
   "dependencies": {
-    "xunit": "2.1.0",
+    "dotnet-test-xunit": "1.0.0-*",
     "Microsoft.ApplicationInsights.AspNetCore": {
-        "target": "project"
+      "target": "project"
     },
-    "dotnet-test-xunit": "1.0.0-*",
     "Microsoft.AspNetCore.Hosting": "1.0.0",
     "Microsoft.AspNetCore.Http.Abstractions": "1.0.0",
```
SergeyKanzhelev: I wonder why not all 1.0.0 changed to 1.0.1

test/FunctionalTestUtils/project.json
```diff
@@ -1,18 +1,19 @@
 {
+  "version": "1.0.1",
```
SergeyKanzhelev: Why it wasn't here before and why we don't have it for the main package's `project.json`? How the version of our NuGet is defined?

dnduffy: These are tests and it just defaulted to 1.0.0, I figured I would explicitly add it and set it to 1.0.1.
The tests don't go in the NuGet package.

test/FunctionalTestUtils/project.json
dnduffy: Some haven't been revved yet.
Nothing new in them I guess.

test/FunctionalTestUtils/project.json
SergeyKanzhelev: My question is mainly about the main project.json version. I don't see any explicit versioning there so version comes from somewhere magically. Should we rev it to 1.0.1?

dnduffy: It is there, line 6.

dnduffy: https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/d7e4cf6abe1e8548de485bc1ce34ff2778d3d674/src/Microsoft.ApplicationInsights.AspNetCore/project.json

SergeyKanzhelev: oh, missed it

SergeyKanzhelev:
BTW, it is a good idea to start a CHANGELOG.md as we use in other repos so we can write release notes easier

dnduffy:
Okay

test/FunctionalTestUtils/project.json
kartang: I am sorry for the late commenting. We don't need any versioning for tests. And tests do not get the package from nuget, as we specify project references in global.json file.

PR #265, opened 2016-09-17
Update the DNX migration link

mortonfox:


dnduffy:
It's good to add the link to the DNX migration instructions, can you please also keep the other link as well since it has useful links for different migrations.

mortonfox:
Sure. The other link is back in place.

mortonfox:
Actually, I noticed that my initial PR updated the wrong link. My mistake. I'll close this one and start over.

PR #281, opened 2016-11-17
[Work in progress] Refactor AspNetCore AppInsigths package

pakrym:


src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -68,25 +60,24 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
...
-            services.AddScoped<RequestTelemetry>((svcs) => {
-                // Default constructor need to be used
-                var rt = new RequestTelemetry();
-                return rt;
-            });
-            
+            services.AddSingleton<ApplicationInsightsInitializer, ApplicationInsightsInitializer>();
+            services.AddSingleton<IApplicationInsightDiagnosticListener, AspNetCoreHostingDiagnosticListener>();
+            services.AddSingleton<IApplicationInsightDiagnosticListener, AspNetCoreMvcDiagnosticsListener>();
+            services.AddSingleton<IStartupFilter, ApplicationInsightsStartupFilter>();
```
pakrym: AspNetCoreHostingDiagnosticListener injects `TelemetryClient` that injects `TelemetryConfiguration` that requires `IOptions` infrastructure to run and initialize.

NuGet.config
```diff
@@ -2,6 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
+    <add key="local" value="d:\localnuget" />
```
dnduffy: We should remove this from the PR since it won't apply for anyone else.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/AspNetCoreHostingDiagnosticListener.cs
```diff
@@ -0,0 +1,103 @@
...
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DiagnosticAdapter;
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+
+namespace Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners
+{
+    internal class AspNetCoreHostingDiagnosticListener: IApplicationInsightDiagnosticListener
+    {
+        private readonly TelemetryClient _client;
```
dnduffy: nitpick: Sorry to be pedantic, I did ask in the meeting a couple weeks ago that the new code have doc comments and be StyleCop (defaults) compliant.  I would really appreciate that, thank-you. :)

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/AspNetCoreHostingDiagnosticListener.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using Microsoft.ApplicationInsights.AspNetCore.Extensions;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DiagnosticAdapter;
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+
+namespace Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners
+{
+    internal class AspNetCoreHostingDiagnosticListener: IApplicationInsightDiagnosticListener
```
dnduffy: Having "AspNetCore" in the name is redundant to the namespace.

davidfowl: nit: formatting

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
SergeyKanzhelev: Can you pls add a comment in code that explains exactly this

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/AspNetCoreHostingDiagnosticListener.cs
pakrym: Fixed all issues and added rules file (btw. `develop` branch had 144 violations)

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        [DiagnosticName("Microsoft.AspNetCore.Hosting.EndRequest")]
+        public void OnEndRequest(HttpContext httpContext, long timestamp)
+        {
+            if (this.client.IsEnabled())
+            {
+                var telemetry = httpContext.Features.Get<RequestTelemetry>();
+
+                telemetry.Duration = new TimeSpan(timestamp - this.beginRequestTimestamp.Value);
```
dnduffy: We've had some issue with the accuracy of this, please add a TODO comment to use the StopWatch from/like the base SDK.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/AspNetCoreMvcDiagnosticsListener.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+        }
+
+        private string GetNameFromRouteContext(IRouteData routeData)
+        {
+            string name = null;
+
+            if (routeData.Values.Count > 0)
+            {
+                var routeValues = routeData.Values;
```
dnduffy: nitpick: prefer not to use var "ever" but especially not where the type cannot be inferred from the right hand.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/AspNetCoreMvcDiagnosticsListener.cs
```diff
@@ -0,0 +1,86 @@
...
+            var telemetry = httpContext.Features.Get<RequestTelemetry>();
+
+            if (telemetry != null && string.IsNullOrEmpty(telemetry.Name))
+            {
+                string name = this.GetNameFromRouteContext(routeData);
+
+                if (!string.IsNullOrEmpty(name))
+                {
+                    name = httpContext.Request.Method + " " + name;
+                    telemetry.Name = name;
```
dnduffy: @SergeyKanzhelev Does this value for name match with the schema changes discussed in [https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/202](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/202)?

pakrym:
@dnduffy @SergeyKanzhelev  Any other concerns that I can address?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,70 @@
...
+        public ApplicationInsightsLogger(string name, TelemetryClient telemetryClient, LogLevel minimumLevel)
+        {
+            this.categoryName = name;
+            this.telemetryClient = telemetryClient;
+            this.minimumLevel = minimumLevel;
+        }
+
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            return null;
```
cwe1ss: Will this get implemented? User-supplied scopes would be important IMO. 
However, this would also include all request-scopes from aspnet/hosting which is already covered by the DiagnosticListener so this might lead to duplicated information?! Should they be filtered out explicitly?

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
```diff
@@ -1,29 +1,60 @@
...
     {
         /// <summary>
+        /// Initializes a new instance of the <see cref="ApplicationInsightsServiceOptions" /> class.
         /// Application Insights service options that controlls the default behavior of application insights features.
         /// </summary>
         public ApplicationInsightsServiceOptions()
         {
             this.EnableQuickPulseMetricStream = true;
             this.EnableAdaptiveSampling = true;
+            this.LoggerMinimumLevel = LogLevel.Information;
```
dnduffy: Isn't LogLevel.Warning the default when not specified?

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                options.TelemetryChannel.EndpointAddress = this.applicationInsightsServiceOptions.EndpointAddress;
+            }
+
+            foreach (var initializer in this.initializers)
+            {
+                options.TelemetryInitializers.Add(initializer);
+            }
+
+            foreach (var module in this.modules)
```
dnduffy: nitpick: There is no way to tell the type here or above, I don't know what var module or var initializer are when reading the source text in review unless I spend extra time scrolling back up to look for a definition like IEnumerable\<ITelemetryInitializer\>.  I know the "var" debate is a hot topic but my point is to increase readability, especially considering the time pressure we are all under to get this in.

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -0,0 +1,102 @@
...
+                    QuickPulseTelemetryProcessor processor = null;
+                    configuration.TelemetryProcessorChainBuilder.Use((next) => {
+                        processor = new QuickPulseTelemetryProcessor(next);
+                        quickPulseModule.RegisterTelemetryProcessor(processor);
+                        return processor;
+                    });
+                }
+
+                // Enabling Adaptive Sampling and initializing server telemetry channel with configuration
+                if (configuration.TelemetryChannel.GetType() == typeof(ServerTelemetryChannel))
```
dnduffy: Aren't you already in this block because of the check above for "configuration.TelemetryChannel is ServerTelemetryChannel"?  Why check the type again a different way?

dnduffy: Actually I think this was just copied from the old code so sorry that's not your doing.
@SergeyKanzhelev Any idea why we originally had this second conditional for ServerTelemetryChannel inside the outer one?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Microsoft.ApplicationInsights.AspNetCore.Logging
+{
+    using Microsoft.Extensions.Logging;
+
+    internal class ApplicationInsightsLoggerProvider : ILoggerProvider
```
dnduffy: Why is all the ILogger stuff internal?
Shouldn't a customer be able to use it if so desired?
This is something we had on our roadmap to consider.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
```diff
@@ -11,7 +10,7 @@
     /// <summary>
     /// A telemetry initializer that will gather Azure Web App Role Environment context information.
     /// </summary>
-    public class AzureWebAppRoleEnvironmentTelemetryInitializer : TelemetryInitializerBase
+    internal class AzureWebAppRoleEnvironmentTelemetryInitializer : TelemetryInitializerBase
```
dnduffy: I've noticed a lot of things being marked as internal (and the friend assembly reference for tests), any particular reason why?

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/OperationNameTelemetryInitializer.cs
```diff
@@ -40,74 +24,9 @@ protected override void OnInitializeTelemetry(HttpContext platformContext, Reque
                 {
                     // We didn't get BeforeAction notification
                     string name = platformContext.Request.Method + " " + platformContext.Request.Path.Value;
-                    requestTelemetry.Name = name;
```
dnduffy: Why is this gone?  (I'm sure there's a good reason I just don't have intellisense or anything in the review to investigate).

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/OperationNameTelemetryInitializer.cs
```diff
@@ -40,74 +24,9 @@ protected override void OnInitializeTelemetry(HttpContext platformContext, Reque
...
-                        {
-                            string arguments = string.Join(@"/", sortedKeys);
-                            name += " [" + arguments + "]";
-                        }
-                    }
-                }
-            }
-
-            return name;
-        }
```
dnduffy: This looks like it just moved to MvcDiagnosticsListener, is that correct?

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -30,10 +31,12 @@
             "version": "1.2.0-beta2",
             "type": "build"
         },
-        "Microsoft.ApplicationInsights": "2.2.0-beta3",
-        "Microsoft.AspNetCore.Hosting.Abstractions": "1.0.0",
-        "Microsoft.AspNetCore.Http.Abstractions": "1.0.0",
-        "Microsoft.AspNetCore.Mvc.ViewFeatures": "1.0.1",
+        "Microsoft.ApplicationInsights": "2.2.0-beta4-*",
```
dnduffy: beta4 has shipped so the * can be removed.

test/EmptyApp.FunctionalTests/project.json
```diff
@@ -4,23 +4,18 @@
...
+    "nowarn": [ "CS0618" ]
   },
   "dependencies": {
-    "dotnet-test-xunit": "1.0.0-*",
+    "dotnet-test-xunit": "2.2.0-*",
     "FunctionalTestUtils": { "target": "project" },
     "Microsoft.AspNetCore.Diagnostics": "1.0.0",
-    "Microsoft.NETCore.Platforms": "1.0.1",
-    "xunit": "2.1.0"
+    "xunit": "2.2.0-*"
```
dnduffy: TODO: We're going to have to file for OSS use approval for this version...

test/FunctionalTestUtils/TelemetryTestsBase.cs
```diff
@@ -27,19 +29,24 @@ public void ValidateBasicRequest(InProcessServer server, string requestPath, Req
             var httpClientHandler = new HttpClientHandler();
             httpClientHandler.UseDefaultCredentials = true;
 
-            var httpClient = new HttpClient(httpClientHandler, true);
-            var task = httpClient.GetAsync(server.BaseHost + requestPath);
-            task.Wait(TestTimeoutMs);
+            Task<HttpResponseMessage> task;
+            using (var httpClient = new HttpClient(httpClientHandler, true))
```
dnduffy: Thank-you for fixing stuff like this, clearly I need to add an FxCop pass to my TODO list for this SDK.

test/MVCFramework45.FunctionalTests/project.json
```diff
@@ -14,11 +14,12 @@
         "wwwroot",
         "Views"
       ]
-    }
+    },
+    "nowarn": [ "CS0618" ]
```
dnduffy: Why is the warning for using obsolete members being turned off (here an elsewhere)?

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
pakrym: I chose the same default logging level ConsoleLogger has:
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerFactoryExtensions.cs#L28

src/Microsoft.ApplicationInsights.AspNetCore/ExceptionTrackingMiddleware.cs
```diff
@@ -1,42 +0,0 @@
...
-    using System;
-    using System.Threading.Tasks;
-    using Microsoft.ApplicationInsights.DataContracts;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation;
-    using Microsoft.AspNetCore.Http;
-
-    /// <summary>
-    /// Sends telemetry about exceptions thrown by the application to the Microsoft Application Insights service.
-    /// </summary>
-    public sealed class ExceptionTrackingMiddleware
```
pakrym: Unhandled exception would still be reported by https://github.com/aspnet/Hosting/blob/b6da89f54cff11474f17486cdc55c2f21f2bbd6b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs#L79

test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/ComponentVersionTelemetryInitializerTests.cs
```diff
@@ -48,19 +47,17 @@ public void InitializeDoesNotOverrideExistingVersion()
...
-            var initializer = new ComponentVersionTelemetryInitializer(config);
+            var initializer = new ComponentVersionTelemetryInitializer(this.BuildConfigurationWithVersion(null));
             var telemetry = new RequestTelemetry();
             initializer.Initialize(telemetry);
         }
 
-        private IConfigurationRoot BuildConfigurationWithVersion()
+        private IOptions<ApplicationInsightsServiceOptions> BuildConfigurationWithVersion(string versions = "1.0.0")
         {
-            Environment.SetEnvironmentVariable(VersionKey, "1.0.0");
```
dnduffy: @SergeyKanzhelev Do you know what the significance of this environment variable setting was?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerProvider.cs
pakrym: Logger gest added into `ILoggerFactory` in `ApplicationInsightsInitializer` so there is no need for user to have access to classes to be able to log.

test/MVCFramework45.FunctionalTests/project.json
pakrym: We are using tons of obsolete API's from base sdk, this is just to clean up build output from things we are ignoring anyway.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -100,13 +100,10 @@
...
         "netstandard1.6": {
-            "imports": [
-                "dotnet5.6",
-                "portable-dnxcore50+net451+net45+win8+wp8+wpa81"
-            ],
-            "dependencies": {
-                "System.Net.NameResolution": "4.0.0"
-            }
+          "dependencies": {
+            "NETStandard.Library": "1.6.0",
```
pakrym: No, if base sdk is not netstandard1.5 this would not work.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/OperationNameTelemetryInitializer.cs
pakrym: Yes, I've split this class into initializer and diagnostics listener.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/OperationNameTelemetryInitializer.cs
pakrym: This was moved to `HostingDiagnosticListener.OnEndRequest` method.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureWebAppRoleEnvironmentTelemetryInitializer.cs
pakrym: My idea was to only expose API's that we are sure about in this release, so if we decide to rearrange things internally that would not count as another breaking change.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsWebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// Configures <see cref="IWebHostBuilder"/> to use Application Insights services.
+        /// </summary>
+        /// <param name="webHostBuilder">The <see cref="IWebHostBuilder"/> instance.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder UseApplicationInsights(this IWebHostBuilder webHostBuilder)
+        {
+            webHostBuilder.ConfigureServices(collection =>
+            {
+                collection.AddApplicationInsightsTelemetry((Action<ApplicationInsightsServiceOptions>)null);
```
davidfowl: This looks pretty bad. Add an overload that doesn't have the arg.

test/MVCFramework45.FunctionalTests/project.json
dnduffy: Okay, we need those warnings on if we are going to know to fix them when we get back to working on this SDK.  There have been a bunch of schema changes recently in the base SDK and with those warnings turned off we probably won't notice to fix them here.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
dnduffy: Okay, leave the * in the 2.2.0-beta4-* for now then since we'll need to reference the build of the base I'll make today in order to get that.
I'll see about just calling the base SDK update for netstandard1.5 beta5 so there is not a problem with old private debug builds of beta4 that won't have it getting referenced.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
dnduffy: Unless I'm misunderstanding where and how this will get used (quite probable) I'm not sure it should default the same as the console logger, I need to understand where the new ILogger implementation will be used from, can you help point me in the right direction?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLogger.cs
dnduffy: I'm okay with that for now, I think I ignored it initially also in my ILogger prototype that we didn't ship.

src/Microsoft.ApplicationInsights.AspNetCore/ExceptionTrackingMiddleware.cs
dnduffy: Just to be clear I understand, that will get reported to AI?  Can you give me a quick tip on what the path is, once everything is wired up, so that this data gets sent to AI when handled in the framework (since the framework doesn't know about AI as such)?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerProvider.cs
pakrym: User would need to call `AddApplicationInsights` on `WebHostBuilder/ServiceCollection` this would add logger into `ILoggerFactory` and the next time he resolves `ILogger<T>` application insights would already be connected.

src/Microsoft.ApplicationInsights.AspNetCore/ExceptionTrackingMiddleware.cs
pakrym: Exception gets logged into `Microsoft.AspNetCore.Hosting.UnhandledException` diagnostic source and received by https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/281/files/b650efa55b39cb04fb32c6874812a6f0ba3d08ed#diff-127c7a3a6227c377b206b4904744707cR82 diagnostic listener.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/MvcDiagnosticsListener.cs
```diff
@@ -0,0 +1,86 @@
...
+        private string GetNameFromRouteContext(IRouteData routeData)
+        {
+            string name = null;
+
+            if (routeData.Values.Count > 0)
+            {
+                var routeValues = routeData.Values;
+
+                object controller;
+                routeValues.TryGetValue("controller", out controller);
```
SergeyKanzhelev: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/282

test/EmptyApp.FunctionalTests/project.json
SergeyKanzhelev: I think for xunit it is just a notification, not an approval. But still needs to be filed. @dnduffy will you pls do it?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/MvcDiagnosticsListener.cs
```diff
@@ -0,0 +1,86 @@
...
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.Extensions.DiagnosticAdapter;
+
+    public class MvcDiagnosticsListener : IApplicationInsightDiagnosticListener
+    {
+        internal const string BeforeActionNotificationName = "Microsoft.AspNetCore.Mvc.BeforeAction";
+
+        public string ListenerName { get; } = "Microsoft.AspNetCore";
+
+        [DiagnosticName(BeforeActionNotificationName)]
```
SergeyKanzhelev: I know it was copied. But to make it looks consistent with other diagnostic source callbacks - no need to use a constant here.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/MvcDiagnosticsListener.cs
SergeyKanzhelev: Is there an access to route template? Can/should we use it instead?

src/Microsoft.ApplicationInsights.AspNetCore/ExceptionTrackingMiddleware.cs
dnduffy: Thank-you. :)

SergeyKanzhelev: As a minimum let's keep it as "Obsolete". However I'd recommend to keep it as is in case customer has own error processing middleware and neither unhandled exception nor diagnostics middleware will catch it

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -2,28 +2,29 @@
...
+    using Microsoft.ApplicationInsights.DependencyCollector;
+    using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector;
 #endif
 
+    /// <summary>
+    /// Extension methods for <see cref="IServiceCollection"/> that allow adding Application Insights services to application.
+    /// </summary>
     public static class ApplicationInsightsExtensions
     {
+        private const string VersionKeyFromConfig = "version";
```
SergeyKanzhelev: Curious - is it the right way to take a version of the app?

src/Microsoft.ApplicationInsights.AspNetCore/RequestTrackingMiddleware.cs
```diff
@@ -1,64 +0,0 @@
...
-    using Microsoft.ApplicationInsights.AspNetCore.Extensions;
-    using Microsoft.ApplicationInsights.DataContracts;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation;
-    using Microsoft.AspNetCore.Builder;
-    using Microsoft.AspNetCore.Http;
-
-    /// <summary>
-    /// Sends telemetry about requests handled by the application to the Microsoft Application Insights service.
-    /// </summary>
-    public sealed class RequestTrackingMiddleware
```
SergeyKanzhelev: Let's keep an empty implementation of it with the `[Obsolete]` attribute. We will remove it by the stable version release than

SergeyKanzhelev:
So this PR doesn't have HttpClient listener, correct?

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
pakrym: By default we set it to entry assembly version(https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/281/files/b650efa55b39cb04fb32c6874812a6f0ba3d08ed#diff-3195debb9b9d179b697b76ff29f8a286R20), but still alow reading from config like the previous version did.

pakrym:
@SergeyKanzhelev yes, you are correct. I think sql and httpclient implementations should be contributed to `netstandard` version of `Microsoft.AI.DependencyCollector` because desktop implementations are there already.

src/Microsoft.ApplicationInsights.AspNetCore/RequestTrackingMiddleware.cs
pakrym: Do you want to keep a class or just an empty extension method on `IApplicationBuilder`, because that's what most of the customers used.

SergeyKanzhelev: You are right - just an extension method will be enough

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
pakrym: I already did, `ApplicationInsightsServiceOptions` is constructed with entry assembly version as and `ApplicationVersion` but we are keeping compatibility with previous SDK and allow overriding via config entry.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
pakrym: When you resolve `ILogger` it goes over all loggerfactories registeged (AI logger factory would be one of them) and creates a logger saving it to array (https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging/Logger.cs#L27). `ILogger.Log` would go trough all the loggers and call message to them, so AI logger would get a message and write it to telemetry client.

test/EmptyApp.FunctionalTests/Startup.cs
```diff
@@ -1,30 +1,22 @@
...
-    using Microsoft.AspNetCore.Diagnostics;
-    using Microsoft.AspNetCore.Http.Internal;
-    using Microsoft.AspNetCore.Mvc.Infrastructure;
 
     public class Startup
     {
         // For more information on how to configure your application, visit http://go.microsoft.com/fwlink/?LinkID=398940
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
```
davidfowl: Do you need to explicitly add this here?

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
dnduffy: How does a user control whether or not they want the AI Logger?
How do we control which messages get sent to AI so that users don't accidentally send a lot of data they didn't expect but have to pay for?

test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/ComponentVersionTelemetryInitializerTests.cs
dnduffy: After discussing with Sergey we don't think it was relevant and is therefore safe to delete (plus this is test code).

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
cwe1ss: Does this play nice with the Microsoft.Extensions.Logging.Filter package? IMO setting just the LogLevel is almost never what you want - especially for remote loggers. People use ILogger for their own namespaces so it should be possible to configure separate log levels. 
If it doesn't work with the Filter package, how about extending it to accept a filter func, similar to the other existing loggers in M.E.Logging

test/EmptyApp.FunctionalTests/Startup.cs
davidfowl: Shouldn't this be in the UseApplicationInsights call?

pakrym: Oh, my bad, it is there, this is not required.

pakrym:
@dnduffy I think I've done all things we have discussed. Please check when you have time.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -32,29 +35,72 @@ public static class ApplicationInsightsExtensions
...
+        /// Adds Application Insights services into service collection.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> insance.</param>
+        /// <param name="options">The action used to configure the options.</param>
+        /// <returns>
+        /// The <see cref="IServiceCollection"/>.
+        /// </returns>
+        public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCollection services)
+        {
+            services.TryAddSingleton(typeof(IHttpContextAccessor), typeof(HttpContextAccessor));
```
davidfowl: typeof? Use the generic overload?

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -30,10 +30,12 @@
             "version": "1.2.0-beta2",
             "type": "build"
         },
-        "Microsoft.ApplicationInsights": "2.2.0-beta3",
-        "Microsoft.AspNetCore.Hosting.Abstractions": "1.0.0",
-        "Microsoft.AspNetCore.Http.Abstractions": "1.0.0",
-        "Microsoft.AspNetCore.Mvc.ViewFeatures": "1.0.1",
+        "Microsoft.ApplicationInsights": "2.2.0-beta5",
+        "Microsoft.AspNetCore.Hosting": "1.0.0",
```
davidfowl: You should only need Hosting.Abstractions, not hosting.

davidfowl:
@pakrym did you add an end to end test that uses the new UseApplicatonInsights() extension method on IWebHostBuilder?

PR #287, opened 2016-12-01
Fix tests and remove obsolete fields

kartang:


test/EmptyApp.FunctionalTests/FunctionalTest/TelemetryModuleWorkingEmptyAppTests.cs
```diff
@@ -7,19 +7,20 @@ public class TelemetryModuleWorkingEmptyAppTests : TelemetryTestsBase
     {
         private const string assemblyName = "EmptyApp.FunctionalTests";
 
-#if NET451
-
         [Fact]
         public void TestBasicDependencyPropertiesAfterRequestingBasicPage()
         {
+#if NET451
```
dnduffy: Now it looks like there is a test when there actually is nothing inside it for NetStandard...

kartang: I am not sure why, but visual studio had trouble recognizing these tests. It was not until I moved them in, that the test explorer identified these.

dnduffy: Interesting, well, maybe add a comment to that effect and then go ahead and complete the PR.

PR #292, opened 2016-12-08
Fixed intermittent test failures and updated package references.

dnduffy:
Removed utils project from list of projects to run dotnet test on.
Updated the version of AI SDK's referenced.
Replaced the InRange test with a custom implementation.
Got test start time from the stopwatch instead of DateTimeOffset.Now which is less accurate.

### DIFF
--- a/dirs.proj
+++ b/dirs.proj
@@ -13,7 +13,6 @@
         <!--<Solution Include="src\Microsoft.ApplicationInsights.AspNetCore\Microsoft.ApplicationInsights.AspNetCore.xproj" />-->
         <!-- Tests -->
         <TestProject Condition="$(RunTests) != '' And $(RunTests)" Include="test\Microsoft.ApplicationInsights.AspNetCore.Tests\ " />
-        <TestProject Condition="$(RunTests) != '' And $(RunTests)" Include="test\FunctionalTestUtils\ " />
         <TestProject Condition="$(RunTests) != '' And $(RunTests)" Include="test\WebApiShimFw46.FunctionalTests\ " />
         <TestProject Condition="$(RunTests) != '' And $(RunTests)" Include="test\EmptyApp.FunctionalTests\ " />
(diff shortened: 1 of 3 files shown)


test/FunctionalTestUtils/TelemetryTestsBase.cs
```diff
@@ -45,7 +45,7 @@ public void ValidateBasicRequest(InProcessServer server, string requestPath, Req
             Assert.Equal(expected.Name, actual.Name);
             Assert.Equal(expected.Success, actual.Success);
             Assert.Equal(expected.Url, actual.Url);
-            Assert.InRange(actual.Timestamp, testStart, DateTimeOffset.Now);
+            InRange(actual.Timestamp, testStart, DateTimeOffset.Now);
```
cijothomas: if you changed to high precision clock for testStart, why not use the same high precision clock for endClock as well ?

dirs.proj
```diff
@@ -13,7 +13,6 @@
         <!--<Solution Include="src\Microsoft.ApplicationInsights.AspNetCore\Microsoft.ApplicationInsights.AspNetCore.xproj" />-->
         <!-- Tests -->
         <TestProject Condition="$(RunTests) != '' And $(RunTests)" Include="test\Microsoft.ApplicationInsights.AspNetCore.Tests\ " />
-        <TestProject Condition="$(RunTests) != '' And $(RunTests)" Include="test\FunctionalTestUtils\ " />
```
cijothomas: why this is removed?

test/FunctionalTestUtils/TelemetryTestsBase.cs
dnduffy: Sure, why not, I was just focused on getting it working.

dirs.proj
dnduffy: Because it isn't an actual test project itself but rather the utils that the other test projects use for their tests.

PR #294, opened 2016-12-09
Trying to prevent the weird test timing issues.

dnduffy:
Added xunit runner config file to prevent test parallelization.
Changed formatting, cleared the buffer and checked for the right named object, and added locking.
Subtracted 50 milliseconds from the test start time to compensate for the build lab reporting earlier timestamps for telemetry items.

### DIFF
--- a/test/FunctionalTestUtils/TelemetryTestsBase.cs
+++ b/test/FunctionalTestUtils/TelemetryTestsBase.cs
@@ -5,6 +5,7 @@
     using System.Linq;
     using System.Net.Http;
     using System.Reflection;
+    using System.Runtime.CompilerServices;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.ApplicationInsights.DataContracts;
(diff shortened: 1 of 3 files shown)


test/FunctionalTestUtils/TelemetryTestsBase.cs
```diff
@@ -20,38 +21,45 @@
...
 
+        [MethodImpl(MethodImplOptions.NoOptimization)]
         public void ValidateBasicRequest(InProcessServer server, string requestPath, RequestTelemetry expected)
         {
-            DateTimeOffset testStart = new DateTimeOffset(Stopwatch.GetTimestamp(), TimeSpan.Zero);
-            var timer = Stopwatch.StartNew();
+            lock (noParallelism)
+            {
+                // Subtract 50 milliseconds to hack around strange behavior on build server where the RequestTelemetry.Timestamp is somehow sometimes earlier than now by a few milliseconds.
+                expected.Timestamp = DateTimeOffset.Now.Subtract(TimeSpan.FromMilliseconds(50));
```
dnduffy: I have not been able to figure out why sometimes the timestamp on the telemetry item is a few milliseconds earlier than the timestamp of when the test starts.  I have seen that sometimes there is more than object in the buffer despite my efforts to prevent this from happening.  If anyone has a better idea of how to fix this properly please let me know.

PR #297, opened 2016-12-12
Add EntityFramework diagnostics source

pakrym:
Also reference individual packages instead `AspNetCore.Hosting` as per comment in my previous PR.

@dnduffy

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Diagnostics;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.Extensions.DiagnosticAdapter;
+
+namespace Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners.Implementation
+{
+    /// <summary>
+    /// <see cref="IApplicationInsightDiagnosticListener"/> implementation that listens for evens specific to EntiryFrameworkCore
+    /// </summary>
+    public class EntityFrameworkDiagnosticListener : IApplicationInsightDiagnosticListener
+    {
+        private readonly TelemetryClient client;
+        private readonly ContextData<long> beginDependencyTimestamp = new ContextData<long>();
+
+        public string ListenerName { get; } = "Microsoft.EntityFrameworkCore";
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="T:HostingDiagnosticListener"/> class.
+        /// </summary>
+        /// <param name="client"><see cref="TelemetryClient"/> to post traces to.</param>
+        public EntityFrameworkDiagnosticListener(TelemetryClient client)
+        {
+            this.client = client;
+        }
+
+        [DiagnosticName("Microsoft.EntityFrameworkCore.BeforeExecuteCommand")]
+        public void OnBeginCommand(long timestamp)
+        {
+            beginDependencyTimestamp.Value = timestamp;
+        }
+
+        [DiagnosticName("Microsoft.EntityFrameworkCore.AfterExecuteCommand")]
+        public void OnEndCommand(IDbCommand command, string executeMethod, Guid instanceId, long timestamp, bool isAsync)
+        {
+            var start = beginDependencyTimestamp.Value;
+            var end = timestamp;
+
+            var telemetry = new DependencyTelemetry();
+            telemetry.Name = command.Connection.Database;
+            telemetry.Data = command.CommandText;
+            telemetry.Duration = new TimeSpan(end - start);
+            telemetry.Timestamp = DateTimeOffset.Now - telemetry.Duration;
+            telemetry.Target = command.Connection.Database;
+            telemetry.Type = "SQL";
+            client.TrackDependency(telemetry);
+        }
+
+        /// <summary>
+        /// Proxy interface for <c>IDbCommand</c> from System.Data
+        /// </summary>
+        public interface IDbCommand
+        {
+            string CommandText { get; }
+
+            IDbConnection Connection { get; }
+        }
+
+        /// <summary>
+        /// Proxy interface for <c>IDbConnection</c> from System.Data
+        /// </summary>
+        public interface IDbConnection
+        {
+            string Database { get; }
+        }
+    }
+
+}
--- a/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
@@ -5,13 +5,14 @@
     using Microsoft.Extensions.DiagnosticAdapter;
 
     /// <summary>
-    /// <see cref="IApplicationInsightDiagnosticListener"/> implementation that listens for evens specific to EntiryFrameworkCore
+    /// <see cref="IApplicationInsightDiagnosticListener"/> implementation that listens for events specific to EntiryFrameworkCore.
     /// </summary>
     public class EntityFrameworkDiagnosticListener : IApplicationInsightDiagnosticListener
     {
         private readonly TelemetryClient client;
         private readonly ContextData<long> beginDependencyTimestamp = new ContextData<long>();
 
+        /// <inheritdoc />
         public string ListenerName { get; } = "Microsoft.EntityFrameworkCore";
 
         /// <summary>
@@ -23,12 +24,18 @@
             this.client = client;
         }
 
+        /// <summary>
+        /// Diagnostic event handler method for 'Microsoft.EntityFrameworkCore.BeforeExecuteCommand' event
+        /// </summary>
         [DiagnosticName("Microsoft.EntityFrameworkCore.BeforeExecuteCommand")]
         public void OnBeginCommand(long timestamp)
         {
             beginDependencyTimestamp.Value = timestamp;
         }
 
+        /// <summary>
+        /// Diagnostic event handler method for 'Microsoft.EntityFrameworkCore.AfterExecuteCommand' event
+        /// </summary>
         [DiagnosticName("Microsoft.EntityFrameworkCore.AfterExecuteCommand")]
         public void OnEndCommand(IDbCommand command, string executeMethod, Guid instanceId, long timestamp, bool isAsync)
         {
--- a/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
@@ -1,10 +1,9 @@
-﻿using System;
-using System.Diagnostics;
-using Microsoft.ApplicationInsights.DataContracts;
-using Microsoft.Extensions.DiagnosticAdapter;
-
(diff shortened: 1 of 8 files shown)


SergeyKanzhelev:
Last we discussed this should go to [Dependency collector](https://www.nuget.org/packages/Microsoft.ApplicationInsights.DependencyCollector/) nugget. What changed?

pakrym:
@SergeyKanzhelev we were discussing SqlClient tracing, this diagnostic source can be used with any database client that has entity framework adapter.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -0,0 +1,67 @@
+﻿namespace Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners.Implementation
+{
+    using System;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.Extensions.DiagnosticAdapter;
+
+    /// <summary>
+    /// <see cref="IApplicationInsightDiagnosticListener"/> implementation that listens for evens specific to EntiryFrameworkCore
```
dnduffy: evens = events

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -0,0 +1,67 @@
...
+        public void OnEndCommand(IDbCommand command, string executeMethod, Guid instanceId, long timestamp, bool isAsync)
+        {
+            var start = beginDependencyTimestamp.Value;
+            var end = timestamp;
+
+            var telemetry = new DependencyTelemetry();
+            telemetry.Name = command.Connection.Database;
+            telemetry.Data = command.CommandText;
+            telemetry.Duration = new TimeSpan(end - start);
+            telemetry.Timestamp = DateTimeOffset.Now - telemetry.Duration;
```
dnduffy: We had a discussion the other day that we should probably go through our code and replace all instances of DateTimeOffset.Now with DateTimeOffset.UtcNow because there happens to be a perf difference between them so maybe this one should be updated now.

src/Microsoft.ApplicationInsights.AspNetCore/Extensibility/Implementation/Tracing/AspNetEventSource.cs
```diff
@@ -3,6 +3,9 @@
 //     Copyright (c) Microsoft Corporation. All rights reserved.
 // </copyright>
 //-----------------------------------------------------------------------
+
+using System.Reflection;
```
dnduffy: Can this using go inside the namespace for consistency?  I don't think it is needed for anything outside the namespace.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Microsoft.Extensions.DependencyInjection
+﻿using Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners.Implementation;
```
dnduffy: Please place inside the namespace sorted with the other usings.

test/MVCFramework45.FunctionalTests/FunctionalTest/EntityFrameworkTelemetryTests.cs
```diff
@@ -0,0 +1,37 @@
...
+            InProcessServer server;
+            using (server = new InProcessServer(assemblyName, InProcessServer.UseApplicationInsights))
+            {
+                using (var httpClient = new HttpClient())
+                {
+                    var task = httpClient.GetAsync(server.BaseHost + "/Home/Contact");
+                    task.Wait(TestTimeoutMs);
+                }
+            }
+            var telemetries = server.BackChannel.Buffer.OfType<DependencyTelemetry>()
```
dnduffy: What is the type here?  I'm guessing it might be "DependencyTelemetry[]" but I'm not sure, please replace var with the known type.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/EntityFrameworkDiagnosticListenerTest.cs
```diff
@@ -0,0 +1,157 @@
...
+                });
+
+            Assert.Equal("Database name", sentTelemetry.Name);
+            Assert.Equal("Command text", sentTelemetry.Data);
+            Assert.Equal("SQL", sentTelemetry.Type);
+            Assert.True(datetime < sentTelemetry.Timestamp);
+            Assert.Equal(new TimeSpan(100000), sentTelemetry.Duration);
+        }
+
+        public class MockDbCommand : System.Data.IDbCommand
```
dnduffy: nitpick: I know it's test code but I would love doc comments anyway please.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
dnduffy: I was just thinking of this one that is new, or you could leave it and I'll get it with all the others, probably safe to leave it.  Sorry, I just commented on it when I saw it but probably I should do it with the others all at once so don't worry about it.

SergeyKanzhelev:
@pakrym - In Application Insights we split telemetry collection logic by types. Typically dependencies will be collected by DependencyCollector package.

I do not understand why we need to collect EntityFramework collection right in AspNet package?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <summary>
+        /// Diagnostic event handler method for 'Microsoft.EntityFrameworkCore.AfterExecuteCommand' event
+        /// </summary>
+        [DiagnosticName("Microsoft.EntityFrameworkCore.AfterExecuteCommand")]
+        public void OnEndCommand(IDbCommand command, string executeMethod, Guid instanceId, long timestamp, bool isAsync)
+        {
+            var start = beginDependencyTimestamp.Value;
+            var end = timestamp;
+
+            var telemetry = new DependencyTelemetry();
```
SergeyKanzhelev: How does it work with SQL dependencies collected by DependencyCollector module? Will we have double counting?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -0,0 +1,74 @@
...
+            telemetry.Name = command.Connection.Database;
```
SergeyKanzhelev: Name should be a name of stored procedure or friendly name of SQL command. Not the database name

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -0,0 +1,74 @@
...
+            telemetry.Data = command.CommandText;
+            telemetry.Duration = new TimeSpan(end - start);
+            telemetry.Timestamp = DateTimeOffset.Now - telemetry.Duration;
+            telemetry.Target = command.Connection.Database;
```
SergeyKanzhelev: Do we have Server name? Please include it into target

pakrym: No, because this is provider agnostic there is no common way to get a server name.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
pakrym: Only for SqlClient on full framework.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
pakrym: From what I see it just puts entire command text in there: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/FrameworkSqlProcessing.cs#L128

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
dnduffy: That would be a problem.

SergeyKanzhelev: Options? Only enable this feature for .NET Core? Double counting is not good

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
SergeyKanzhelev: Framework only sets command text as a name for stored procedures. We have quite strange logic for full framework: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/ProfilerSqlProcessing.cs#L144

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
SergeyKanzhelev: Can we make our best attempt? Like parse connection string or something. We want to light up scenario in UI when we open database information from the application map's SQL node

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
pakrym: This is useful for non SqlClient connection on full framework too. We can try and filter out SqlCommands by type.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
pakrym: So it's, `Server | Database | StoredProcedure` for stored procedures and `Server | Database` for everything else. But `DataSource` unfortunately is SqlClient specific.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/EntityFrameworkDiagnosticListener.cs
pakrym: We can send entire connection string but it might contain secrets. We can parse connection string syntax but we don't know what to search for.

PR #305, opened 2017-01-03
Added AddApplicationInsightsTelemetry overload.

dnduffy:


src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -90,6 +90,30 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
         /// Adds Application Insights services into service collection.
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection"/> insance.</param>
+        /// <param name="options">The options instance used to configure with.</param>
+        /// <returns>
+        /// The <see cref="IServiceCollection"/>.
+        /// </returns>
+        public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCollection services, ApplicationInsightsServiceOptions options)
```
Dmitry-Matveev: Are there some tests for the other overloads, so that we can easily add a test for this overload as well?

dnduffy: Yes, I added a new one for this change and updated the PR.

PR #308, opened 2017-01-05
Fixed unit test failure from shared reference.

dnduffy:


test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/ApplicationInsightsExtensionsTests.cs
```diff
@@ -237,6 +237,7 @@ public static void RegistersTelemetryClientToGetTelemetryConfigurationFromContai
                 IServiceProvider serviceProvider = services.BuildServiceProvider();
                 var configuration = serviceProvider.GetTelemetryConfiguration();
                 configuration.InstrumentationKey = Guid.NewGuid().ToString();
+                ITelemetryChannel oldChannel = configuration.TelemetryChannel;
```
dnduffy: We attribute the tests to prevent them from running in parallel because of shared infrastructure stuff kind of like this.

dnduffy: try finally is a good idea, I'll add that.

Issue #315, opened 2017-01-11
Debug trace logger should no-op if application manually adds the ApplicationTraceLoggerProvider itself


DamianEdwards:
Currently, an application that wishes to use the `ApplicationInsightsLoggerProvider` to redirect log messages to Application Insights trace telemetry needs to disable the debug logger (via `ApplicationInsightsOptions.EnableDebugLogger`) or they'll end up with duplicate trace messages while debugging the application.

The logger added by `ApplicationInsightsOptions.EnableDebugLogger` should coordinate this itself to avoid duplicate messages rather than the application having to do it manually (maybe by using a marker `TelemetryInitializer` to detect when another logger instance has been added and in that case evaluate that instance's configured filter [stored on the marker] and no-op if the filter already applies).

Issue #314, opened 2017-01-11
Don't auto-wire-up ILoggerProvider if an instrumentation key is configured


DamianEdwards:
Currently, an `ILoggerProvider` is wired up automatically to funnel log messages through to Application Insights as `TraceTelemetry` calls when the debugger is attached. If the application is configured with an instrumentation key however, we don't want these traces being sent to the subscription in Azure. We should ensure the logger filter only logs if a key isn't present.

Issue #313, opened 2017-01-11
JS snippet should be empty if TelemetryClient.IsEnabled() is false


DamianEdwards:
See https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/develop/src/Microsoft.ApplicationInsights.AspNetCore/ApplicationInsightsJavaScript.cs#L34

This should also be ensuring that `TelemetryClient.IsEnabled()` returns `true`, along with the key being available.

Issue #312, opened 2017-01-10
Add a TelemetryInitializer that adds the ASP.NET Core environment name as a custom property to all tracked telemetry calls


DamianEdwards:
[Like I do here](https://github.com/aspnet/live.asp.net/blob/935487d7c2a1314d8d3a36430039327c88842aed/src/live.asp.net/Services/EnvironmentTelemetryInitializer.cs). This should just happen automatically.

PR #317, opened 2017-01-11
Fix couple 2.0-beta issues

pakrym:
Add AspNetCore environment initializer. Fixes: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/312
Authenticated user tracking script support. Fixes: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/311
Do not output client javascript if telemetry is disabled. Fixes: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/313
Prevent debug logging if instrumentation key exists. Fixes: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/314
Disable debug logger if other loggers are being added to ILoggerFactory. Fixes: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/315

@SergeyKanzhelev @dnduffy

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/ApplicationInsightsJavaScript.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/ApplicationInsightsJavaScript.cs
@@ -1,7 +1,11 @@
 ﻿namespace Microsoft.ApplicationInsights.AspNetCore
 {
     using System.Globalization;
+    using System.Security.Principal;
+    using Microsoft.ApplicationInsights.AspNetCore.Extensions;
     using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.Extensions.Options;
 
     /// <summary>
     /// This class helps to inject Application Insights JavaScript snippet into application code.
@@ -11,29 +15,54 @@
         /// <summary>JavaScript snippet.</summary>
         private static readonly string Snippet = Resources.JavaScriptSnippet;
 
+        /// <summary>JavaScript authenticated user tracking snippet.</summary>
+        private static readonly string AuthSnippet = Resources.JavaScriptAuthSnippet;
+
         /// <summary>Configuration instance.</summary>
         private TelemetryConfiguration telemetryConfiguration;
+
+        /// <summary> Http context accessor.</summary>
+        private readonly IHttpContextAccessor httpContextAccessor;
+
+        /// <summary> Weather to print authenticated user tracking snippet.</summary>
+        private bool enableAuthSnippet;
 
         /// <summary>
         /// Initializes a new instance of the JavaScriptSnippet class.
         /// </summary>
         /// <param name="telemetryConfiguration">The configuration instance to use.</param>
-        public JavaScriptSnippet(TelemetryConfiguration telemetryConfiguration)
+        /// <param name="serviceOptions">Service options instance to use.</param>
+        /// <param name="httpContextAccessor">Http context accessor to use.</param>
+        public JavaScriptSnippet(TelemetryConfiguration telemetryConfiguration,
+            IOptions<ApplicationInsightsServiceOptions> serviceOptions,
+            IHttpContextAccessor httpContextAccessor)
         {
             this.telemetryConfiguration = telemetryConfiguration;
+            this.httpContextAccessor = httpContextAccessor;
+            this.enableAuthSnippet = serviceOptions.Value.EnableAuthenticationTrackingJavaScript;
         }
 
         /// <summary>
         /// Gets a code snippet with instrumentation key initialized from TelemetryConfiguration.
         /// </summary>
-        /// <returns>JavaScript code snippet with instrumentation key or empty if instrumentation key was not set for the applicaiton.</returns>
+        /// <returns>JavaScript code snippet with instrumentation key or empty if instrumentation key was not set for the application.</returns>
         public string FullScript
         {
             get
             {
-                if (!string.IsNullOrEmpty(this.telemetryConfiguration.InstrumentationKey))
+                if (!this.telemetryConfiguration.DisableTelemetry &&
+                    !string.IsNullOrEmpty(this.telemetryConfiguration.InstrumentationKey))
                 {
-                    return string.Format(CultureInfo.InvariantCulture, Snippet, this.telemetryConfiguration.InstrumentationKey);
+                    string additionalJS = string.Empty;
+                    IIdentity identity = httpContextAccessor?.HttpContext.User?.Identity;
+                    if (enableAuthSnippet &&
+                        identity != null &&
+                        identity.IsAuthenticated)
+                    {
+                        string escapedUserName = identity.Name.Replace("\\", "\\\\");
+                        additionalJS = string.Format(CultureInfo.InvariantCulture, AuthSnippet, escapedUserName);
+                    }
+                    return string.Format(CultureInfo.InvariantCulture, Snippet, this.telemetryConfiguration.InstrumentationKey, additionalJS);
                 }
                 else
                 {
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
@@ -16,6 +16,7 @@
             this.EnableQuickPulseMetricStream = true;
             this.EnableAdaptiveSampling = true;
             this.EnableDebugLogger = true;
+            this.EnableAuthenticationTrackingJavaScript = false;
             this.ApplicationVersion = Assembly.GetEntryAssembly().GetName().Version.ToString();
         }
 
@@ -55,5 +56,11 @@
         /// Gets or sets a value indicating whether a logger would be registered automatically in debug mode.
         /// </summary>
         public bool EnableDebugLogger { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether a JavaScript to track current authenticated user would be printed along with main
+        /// ApplicationInsight tracking JavaScript.
+        /// </summary>
+        public bool EnableAuthenticationTrackingJavaScript { get; set; }
     }
 }
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
@@ -58,8 +58,8 @@
         public bool EnableDebugLogger { get; set; }
 
         /// <summary>
-        /// Gets or sets a value indicating whether a JavaScript to track current authenticated user would be printed along with main
-        /// ApplicationInsight tracking JavaScript.
+        /// Gets or sets a value indicating whether a JavaScript snippet to track the current authenticated user should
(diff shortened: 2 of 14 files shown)


pakrym:
Also fixes null reference exception while determining application version if ran from unit tests: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/310

gardnerjr:
@DamianEdwards  for #314 does this mean that if you have an ikey you don't get any output in the console when debugging?  this would mean that all of the VS tooling for local search/etc would stop working once you've set an ikey.  or am i interpreting this wrong and that output would still go out?

src/Microsoft.ApplicationInsights.AspNetCore/ApplicationInsightsJavaScript.cs
```diff
@@ -11,29 +15,54 @@ public class JavaScriptSnippet
...
-        public JavaScriptSnippet(TelemetryConfiguration telemetryConfiguration)
+        /// <param name="serviceOptions">Service options instance to use.</param>
+        /// <param name="httpContextAccessor">Http context accessor to use.</param>
+        public JavaScriptSnippet(TelemetryConfiguration telemetryConfiguration,
+            IOptions<ApplicationInsightsServiceOptions> serviceOptions,
+            IHttpContextAccessor httpContextAccessor)
         {
             this.telemetryConfiguration = telemetryConfiguration;
+            this.httpContextAccessor = httpContextAccessor;
+            this.enableAuthSnippet = serviceOptions.Value.EnableAuthenticationTrackingJavaScript;
```
dnduffy: Can "Value" ever be null?

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
```diff
@@ -16,7 +16,8 @@ public ApplicationInsightsServiceOptions()
             this.EnableQuickPulseMetricStream = true;
             this.EnableAdaptiveSampling = true;
             this.EnableDebugLogger = true;
-            this.ApplicationVersion = Assembly.GetEntryAssembly().GetName().Version.ToString();
+            this.EnableAuthenticationTrackingJavaScript = false;
+            this.ApplicationVersion = Assembly.GetEntryAssembly()?.GetName().Version.ToString();
```
dnduffy: Just curious but when can GetEntryAssembly return null?

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
```diff
@@ -55,5 +56,11 @@ public ApplicationInsightsServiceOptions()
         /// Gets or sets a value indicating whether a logger would be registered automatically in debug mode.
         /// </summary>
         public bool EnableDebugLogger { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether a JavaScript to track current authenticated user would be printed along with main
```
dnduffy: nitpick grammar: "Gets or sets a value indicating whether a JavaScript snippet to track the current authenticated user should be printed along with the main ApplicationInsights tracking script."

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsServiceOptions.cs
pakrym: When running in xunit test.

pakrym:
@gardnerjr we would only disable automatic debug time logger in that case so you don't accidentally write tons of traces directly to Azure. All other integration features would work as expected.

src/Microsoft.ApplicationInsights.AspNetCore/ApplicationInsightsJavaScript.cs
pakrym: No, I don't think so.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
```diff
@@ -17,24 +17,23 @@ public static class ApplicationInsightsLoggerFactoryExtensions
...
-            return factory.AddApplicationInsights(client, LogLevel.Warning);
+            return factory.AddApplicationInsights(serviceProvider, LogLevel.Warning);
         }
 
         /// <summary>
         /// Adds an ApplicationInsights logger that is enabled for <see cref="LogLevel"/>s of minLevel or higher.
         /// </summary>
         /// <param name="factory"></param>
-        /// <param name="client">The instance of <see cref="TelemetryClient"/> to use for logging.</param>
+        /// <param name="serviceProvider">The instance of <see cref="IServiceProvider"/> to use for logging.</param>
```
dnduffy: The param comment here and below should be updated to reflect the new type being passed since the service provider isn't being used for logging.

src/Microsoft.ApplicationInsights.AspNetCore/ApplicationInsightsJavaScript.cs
DamianEdwards: The `IOptions` infrastructure will ensure a value is created.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
```diff
@@ -43,12 +42,19 @@ public static class ApplicationInsightsLoggerFactoryExtensions
...
             this ILoggerFactory factory,
-            TelemetryClient client,
+            IServiceProvider serviceProvider,
             Func<string, LogLevel, bool> filter)
         {
+            var client = serviceProvider.GetService<TelemetryClient>();
+            var debugLoggerControl = serviceProvider.GetService<DebugLoggerControl>();
+            if (debugLoggerControl != null)
+            {
+                debugLoggerControl.EnableDebugLogger = false;
```
dnduffy: Just curious so I understand this, when DebugLoggerControl is created the flag is defaulted to true but then here when adding App Insights it is set to false, can you explain the flow for enabling and disabling the debug logger?

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AspNetCoreEnvironmentTelemetryInitializer.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AspNetCoreEnvironmentTelemetryInitializer"/> class.
+        /// </summary>
+        public AspNetCoreEnvironmentTelemetryInitializer(IHostingEnvironment environment)
+        {
+            _environment = environment;
+        }
+
+        public void Initialize(ITelemetry telemetry)
```
dnduffy: Please add doc comments.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AspNetCoreEnvironmentTelemetryInitializer.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private const string AspNetCoreEnvironmentPropertyName = "AspNetCoreEnvironment";
+        private readonly IHostingEnvironment _environment;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AspNetCoreEnvironmentTelemetryInitializer"/> class.
+        /// </summary>
+        public AspNetCoreEnvironmentTelemetryInitializer(IHostingEnvironment environment)
+        {
+            _environment = environment;
```
dnduffy: nitpick: prefer StyleCop compliance "this.environment" vs. "_environment".

test/Microsoft.ApplicationInsights.AspNetCore.Tests/JavaScript/ApplicationInsightsJavaScriptTest.cs
```diff
@@ -10,27 +15,93 @@ public static class ApplicationInsightsJavaScriptTest
...
+            string unittestkey = "unittestkey";
             var telemetryConfiguration = new TelemetryConfiguration { InstrumentationKey = unittestkey };
-            var snippet = new JavaScriptSnippet(telemetryConfiguration);
-            Assert.Contains("instrumentationKey: '" + unittestkey + "'", snippet.FullScript.ToString());
+            var snippet = new JavaScriptSnippet(telemetryConfiguration, GetOptions(true), GetHttpContextAccessor("username", true));
+            Assert.Contains("setAuthenticatedUserContext(\"username\")", snippet.FullScript);
+        }
+
+        [Fact]
+        public static void SnippetWillNotIncludeAuthUserNameIfEnabledAndAuthenticated()
```
dnduffy: Other than the "\\" in the username in the next test what is the difference between this test and the one after it?
Why does "\\" specify whether or not someone is authenticated?  Couldn't there be an identity object, with a name which does not contain slashes, that is for an authenticated user from some system other than Active Directory?

dnduffy: Maybe I'm misunderstanding what is happening, does the second parameter to GetHttpContextAccessor specify if a value is returned?  Could you add some doc comments to these test methods to explain exactlky what and how they are testing?  Thank-you.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/JavaScript/ApplicationInsightsJavaScriptTest.cs
```diff
@@ -10,27 +15,93 @@ public static class ApplicationInsightsJavaScriptTest
...
+            return new HttpContextAccessor
+            {
+                HttpContext = new HttpContextStub
+                {
+                    User = new GenericPrincipal(new IdentityStub() { Name = name, IsAuthenticated = isAuthenticated }, new string[0])
+                }
+            };
+        }
+
+        private class IdentityStub : IIdentity
```
dnduffy: Please put doc comments on new code.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
pakrym: Idea here is to enable debug time logger only when there is no other logger added explicitly. So when adding debug logger we include  `debugLoggerControl.EnableDebugLogger` into it's filter statement https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/317/files#diff-ba05ccca641cf8f4dbb9accc9dc93780R37 and when you call `loggetFactory.AddApplicationInsights` `EnableDebugLogger ` get's set to `false` and debug logger disabled.

pakrym:
Done

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
dnduffy: Thank-you.

PR #322, opened 2017-01-19
Added explicit setting of the SDK version to ILogger trace calls.

dnduffy:


dnduffy:
I don't think this actually fixes the problem I was trying to address because I still get telemetry with the base SDK version being reported which doesn't go through this path.  If I recall correctly this logger isn't used except in debug mode with no iKey.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLogger.cs
```diff
@@ -41,19 +45,21 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
                 {
-                    foreach (var item in stateDictionary)
+                    foreach (KeyValuePair<string, object> item in stateDictionary)
                     {
                         dict[item.Key] = Convert.ToString(item.Value);
                     }
                 }
 
-                this.telemetryClient.TrackTrace(formatter(state, exception), this.GetSeverityLevel(logLevel), dict);
+                traceTelemetry.Context.GetInternalContext().SdkVersion = this.sdkVersion;
```
pakrym: Why not have initializer doing this instead?

pakrym:
Tests?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLogger.cs
SergeyKanzhelev: Initializers will set the version for all telemetry item - custom reported, collected by dependency collector module, etc. We want to distinguish those so we can help troubleshoot the issue when we see "bad" trace telemetry. Knowing how that "bad" trace was collected is essential

pakrym: Then I would prefer having this `traceTelemetry.Context.GetInternalContext().SdkVersion = ` and ` this.sdkVersion = SdkVersionUtils.VersionPrefix + SdkVersionUtils.GetAssemblyVersion();` in common utility class. Just don't like increasing code duplication

dnduffy: I'll create an issue to track that cleanup as an enhancement for the next build.

dnduffy:
Unit test added.

PR #329, opened 2017-01-20
Removed UserAgentTelemetryInitializer and associated tests.

dnduffy:


flagbug:
@dnduffy @SergeyKanzhelev I'm wondering, why has this been removed? Having the user agent showing up in the telemetry would be pretty useful!

dnduffy:
@flagbug If you use the JavaScript snippet on your pages to bring in the JS SDK then I believe you will get it from the web page telemetry and it was just a redundant waste of space duplicated from the server side.

joelmalone:
@dnduffy For methods that don't return Javascript, e.g. an API method that returns Json, what's the preferred way to inject User Agent into the AI telemetry?  It would be easy enough to just re-implement the deleted UserAgentTelemetryInitializer in my project, so just wondering if there's a better way.

SergeyKanzhelev:
@joelmalone yes, you can re-implement initializer. Note that you should use custom properties to store the User Agent string value.

Issue #321, opened 2017-01-18
Update HttpContext.TraceIdenfier with Telemetry.Name generated in MvcDiagnosticsListener


northtyphoon:
MvcDiagnosticsListener generates the operation name for request telemetry from IRouteData. 

Application Insights Profiler needs to correlate the profiling data with the request telemetry by operation name. 

Currently Application Insights Profiler subscribes to Microsoft.AspNetCore.Hosting diagnostic source through [DiagnosticSourceEventSource.cs](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs) but it only read the properties from HttpContext . 

To pass the operation name to Application Insights Profiler, MvcDiagnosticsListener can append the name to HttpContext.TraceIdenfier. eg, 

        [DiagnosticName("Microsoft.AspNetCore.Mvc.BeforeAction")]
        public void OnBeforeAction(HttpContext httpContext, IRouteData routeData)
        {
            var telemetry = httpContext.Features.Get<RequestTelemetry>();
            if (telemetry != null && string.IsNullOrEmpty(telemetry.Name))
            {
                string name = this.GetNameFromRouteContext(routeData);
                if (!string.IsNullOrEmpty(name))
                {
                    name = httpContext.Request.Method + " " + name;
                    telemetry.Name = name;
                    httpContext.TraceIdentifier += “|” + name
                }
            }
        }

SergeyKanzhelev:
Would it be better to associate AI `RequestTelemetry` with the `TraceIdentifier`? So if you listen for all `RequestTelemetry` objects using rich event source you can associate trace identifier with the name

northtyphoon:
The additional correlation events based on rich event source have the cost. First, we need to pipe the data to AI profiler process which is complicated in Antares. Second, to correlate the request and profiling trace, we need to keep the data in each profiling cycle for mapping.

In aspnetcore app, we both subscribe to the same aspnetcore diagnostic source and read the data from httpcontext in most cases. So we want to explorer a solution that can avoid the additional correlation event based on rich event source. 

We have a prototype that works based on the single diagnostics source as long as the name can be passed through HttpContext.TraceIdentifier.

@vancem @weng5e

SergeyKanzhelev:
1. Our SDK is trying to be "least" invasive. Attaching the name to trace identifier will not be expected by other telemetry solutions.
[...]

PR #334, opened 2017-01-25
Update HttpContext.TraceIdenfier with Telemetry.Name

northtyphoon:
Update HttpContext.TraceIdentifier with Telemetry.Name generated in MvcDiagnosticsListener.

The motivation of the change is described in #321

### DIFF
--- a/test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
+++ b/test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
@@ -1,12 +1,10 @@
-﻿using Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners;
-
-namespace Microsoft.ApplicationInsights.AspNetCore.Tests.TelemetryInitializers
+﻿namespace Microsoft.ApplicationInsights.AspNetCore.Tests.TelemetryInitializers
 {
     using System;
     using System.Diagnostics;
-    using System.Diagnostics.Tracing;
     using Microsoft.ApplicationInsights.AspNetCore.TelemetryInitializers;
     using Microsoft.ApplicationInsights.AspNetCore.Tests.Helpers;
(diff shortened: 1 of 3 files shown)


test/Microsoft.ApplicationInsights.AspNetCore.Tests/TelemetryInitializers/OperationNameTelemetryInitializerTests.cs
```diff
@@ -66,13 +66,15 @@ public void InitializeDoesNotOverrideOperationNameProvidedInline()
...
             var contextAccessor = HttpContextAccessorHelper.CreateHttpContextAccessor(new RequestTelemetry(), null);
+            string originalTraceIdentifier = contextAccessor.HttpContext.TraceIdentifier;
 
             var initializer = new OperationNameTelemetryInitializer(contextAccessor);
 
             var telemetry = new EventTelemetry();
             initializer.Initialize(telemetry);
 
             Assert.Equal("GET /Test", telemetry.Context.Operation.Name);
+            Assert.Equal(originalTraceIdentifier, contextAccessor.HttpContext.TraceIdentifier);
```
dnduffy: You read out the string value from contextAccessor.HttpContext.TraceIdentifier and then assert that it matches reading it out again?  I don't see this test making a call which would cause the diagnostic listener code to run so asserting that the value of contextAccessor.HttpContext.TraceIdentifier matches itself when nothing has written to it doesn't seem to be very useful.

northtyphoon: Will create new tests for the change

SergeyKanzhelev:
I'm closing this PR as I do not agree with approach:

1. Our SDK is trying to be "least" invasive. Attaching the name to trace identifier will not be expected by other telemetry solutions.
2. We already misuse diagnostics source for http headers injection. We discussed it explicitly and agreed that it is not invasive.
3. Request `name` can be changed during the action execution. So this name may not be a final one.
4. There are other properties besides the `name` that can be used for correlation between SP and AI. So the approach of receiving of entire `RequestTelemetry` is better

I'll duplicate this response in the original issue.

PR #335, opened 2017-01-26
Add InstrumentationKey headers

daschult:
This pull request adds [DependencyCollector](https://github.com/Microsoft/ApplicationInsights-dotnet-server/tree/develop/Src/DependencyCollector) functionality to ASP.NET Core. In addition, this pull request also adds [source and target instrumentation key headers](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/8c51d66c08c1baf0844613637fa5d3ccfc7590f9/Src/Common/RequestResponseHeaders.cs) to the dependency collector and hosting diagnostic listeners.


SergeyKanzhelev:
We discussed that this functionality should go to DependencyCollector NuGet, not to aspnet NuGet. Why have you implemented it here?

Would httpclient use HttpWebRequest on full framework? Will the dependency call be duplicated when you run on full framework?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+
+            this.requestTelemetry.TryAdd(loggingRequestId, telemetry);
+        }
+
+        /// <summary>
+        /// Diagnostic event handler method for 'System.Net.Http.Response' event.
+        /// </summary>
+        [DiagnosticName("System.Net.Http.Response")]
+        public void OnResponseReceived(HttpResponseMessage response, Guid loggingRequestId, long timestamp)
```
pakrym: Are you sure this always gets called and won't leak memory in `requestTelemetry` dictionary?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -38,10 +39,24 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
         {
             if (this.client.IsEnabled())
             {
-                httpContext.Features.Set(new RequestTelemetry());
+                RequestTelemetry requestTelemetry = new RequestTelemetry();
+                if (string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey))
+                {
+                    this.client.Initialize(requestTelemetry);
```
pakrym: Why do it here?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -38,10 +39,24 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
...
+                }
+
+                httpContext.Features.Set(requestTelemetry);
 
                 this.beginRequestTimestamp.Value = timestamp;
                 this.client.Context.Operation.Id = httpContext.TraceIdentifier;
+                requestTelemetry.Id = httpContext.TraceIdentifier;
+
+                IHeaderDictionary responseHeaders = httpContext.Response?.Headers;
+                if (!string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey)
```
pakrym: Extract to initializer?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            string hash;
+            var found = knownIKeyHashes.TryGetValue(instrumentationKey, out hash);
+
+            if (!found)
+            {
+                // Simplistic cleanup to guard against this becoming a memory hog.
+                if (knownIKeyHashes.Keys.Count >= MAXSIZE)
+                {
+                    knownIKeyHashes.Clear();
```
pakrym: 👍

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/InstrumentationKeyHashLookupHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <returns>SHA256 hash for the instrumentation key.</returns>
+        public static string GetInstrumentationKeyHash(string instrumentationKey)
+        {
+            if (string.IsNullOrEmpty(instrumentationKey))
+            {
+                throw new ArgumentNullException("instrumentationKey");
+            }
+
+            string hash;
+            var found = knownIKeyHashes.TryGetValue(instrumentationKey, out hash);
```
pakrym: GetOrAdd?

test/EmptyApp.FunctionalTests/Startup.cs
```diff
@@ -46,6 +47,14 @@ public void Configure(IApplicationBuilder app)
                         telemetryClient.TrackTrace("Fetched contact details.", SeverityLevel.Information);
                         await context.Response.WriteAsync("Hello!");
                     }
+                    else if (context.Request.GetUri().ToString().Contains("Dependency"))
+                    {
+                        using (HttpClient client = new HttpClient())
+                        {
+                            await client.GetAsync("https://www.microsoft.com/");
```
pakrym: Can we hit localhost instead of microsoft.com?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -112,6 +127,25 @@ private void EndRequest(HttpContext httpContext, long timestamp)
                     telemetry.Name = httpContext.Request.Method + " " + httpContext.Request.Path.Value;
                 }
 
+                if (string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
+                {
+                    this.client.Initialize(telemetry);
```
pakrym: Why do we initialize telemetry prematurely and no on write?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,103 @@
...
+            DependencyTelemetry telemetry = new DependencyTelemetry();
+            telemetry.Start();
+            this.client.Initialize(telemetry);
+
+            telemetry.Name = resourceName;
+            telemetry.Target = requestUri.Host;
+            telemetry.Type = "Http";
+            telemetry.Data = requestUri.OriginalString;
+
+            // Add the source instrumentation key header if collection is enabled, the request host is not in the excluded list and the same header doesn't already exist
```
brahmnes: The comment mentions the exclusion list, but I don't see that in the code.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -38,10 +39,24 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
         {
             if (this.client.IsEnabled())
             {
-                httpContext.Features.Set(new RequestTelemetry());
+                RequestTelemetry requestTelemetry = new RequestTelemetry();
```
brahmnes: Isn't the RequestTelemetry object already created somewhere else, since the AI SDK already tracks request today.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        /// <summary>
+        /// Diagnostic event handler method for 'System.Net.Http.Response' event.
+        /// </summary>
+        [DiagnosticName("System.Net.Http.Response")]
+        public void OnResponseReceived(HttpResponseMessage response, Guid loggingRequestId, long timestamp)
+        {
+            DependencyTelemetry telemetry;
+            if (this.requestTelemetry.TryRemove(loggingRequestId, out telemetry))
```
nizarq: What situation might TryRemove return false? If it does, do we need to log something or just silently move on?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                if (response.Headers.Contains(RequestResponseHeaders.TargetInstrumentationKeyHeader))
+                {
+                    string targetInstrumentationKeyHash = response.Headers.GetValues(RequestResponseHeaders.TargetInstrumentationKeyHeader).SingleOrDefault();
+
+                    // We only add the cross component correlation key if the key does not remain the current component.
```
nizarq: does not 'represent'

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -38,10 +39,24 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
...
+
+                httpContext.Features.Set(requestTelemetry);
 
                 this.beginRequestTimestamp.Value = timestamp;
                 this.client.Context.Operation.Id = httpContext.TraceIdentifier;
+                requestTelemetry.Id = httpContext.TraceIdentifier;
+
+                IHeaderDictionary responseHeaders = httpContext.Response?.Headers;
+                if (!string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey)
+                    && !responseHeaders.ContainsKey(RequestResponseHeaders.TargetInstrumentationKeyHeader))
```
nizarq: responseHeaders could be null and may through a null ref here

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
nizarq: At least in .net (full) sdk, this forces context initialization, which means requestTelemetry.Context.InstrumenationKey would be set if the client has the instrumentation key. Since requestTelemetry.Context.InstrumenationKey is used in the method below - I guess this call is here to make sure the value is present before using it.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/DependencyCollectorDiagnosticListenerTest.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        [Fact]
+        public void OnRequestSentAndResponseReceived()
+        {
+            Guid loggingId = new Guid();
+            Uri requestUri = new Uri("http://www.microsoft.com/test/path.html");
+
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, requestUri);
+            listener.OnRequestSent(request, loggingId, 0);
```
nizarq: this.listener here and at other places.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/DependencyCollectorDiagnosticListenerTest.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            Assert.False(request.Headers.Contains(RequestResponseHeaders.TargetInstrumentationKeyHeader));
+
+            // We can't check this header value because it changes every time the test is run.
+            Assert.True(request.Headers.Contains(RequestResponseHeaders.StandardParentIdHeader));
+
+            Assert.False(request.Headers.Contains(RequestResponseHeaders.StandardRootIdHeader));
+        }
+
+        private static void AssertTelemetry(ITelemetry telemetry, Uri requestUri, string expectedTarget = null, string expectedType = null, bool? expectedSuccess = null, string expectedResultCode = null)
```
nizarq: ValidateTelemetry instead of AssertTelemetry??

nizarq: Or AssertTelemetryValid

test/EmptyApp.FunctionalTests/Startup.cs
dnduffy: I think microsoft.com was so widely used by companies as their ping test destination that MSIT turned off ping responses a few years ago.  I realize this is not a ping but the spirit is the same, we shouldn't abuse our network for testing if there is another option.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
pakrym: But there is a client instance in this method, so instrumentation key can be retrieved from client configuration.

test/EmptyApp.FunctionalTests/Startup.cs
pharring: www.example.com

test/Microsoft.ApplicationInsights.AspNetCore.Tests/DependencyCollectorDiagnosticListenerTest.cs
```diff
@@ -0,0 +1,156 @@
...
+            HttpResponseMessage response = new HttpResponseMessage(HttpStatusCode.OK);
+            listener.OnResponseReceived(response, loggingId, 1);
+
+            AssertTelemetry(this.sentTelemetry, requestUri);
+        }
+
+        [Fact]
+        public void OnRequestSentAndResponseReceivedWhereTargetInstrumentationKeyIsSameAsSource()
+        {
+            Guid loggingId = new Guid();
```
pharring: Do you mean ```Guid.NewGuid()``` ? What you wrote creates the equivalent of ```default(Guid)```

test/Microsoft.ApplicationInsights.AspNetCore.Tests/DependencyCollectorDiagnosticListenerTest.cs
```diff
@@ -0,0 +1,156 @@
...
+            response.Headers.Add(RequestResponseHeaders.TargetInstrumentationKeyHeader, request.Headers.GetValues(RequestResponseHeaders.SourceInstrumentationKeyHeader).Single());
+            listener.OnResponseReceived(response, loggingId, 1);
+
+            AssertTelemetry(this.sentTelemetry, requestUri);
+        }
+
+        [Fact]
+        public void OnRequestSentAndResponseReceivedWhereTargetInstrumentationKeyIsNotSameAsSource()
+        {
+            Guid loggingId = new Guid();
```
pharring: ```Guid.NewGuid()``` (and elsewhere in this file)

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -112,6 +127,25 @@ private void EndRequest(HttpContext httpContext, long timestamp)
                     telemetry.Name = httpContext.Request.Method + " " + httpContext.Request.Path.Value;
                 }
 
+                if (string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
```
pharring: Should this (setting the iKey based on request headers) be done in a TelemetryInitializer? That way, any other telemetry associated with the request (e.g. Exceptions, Event, etc.) sent prior to OnEndRequest will also be tagged with the same iKey. Or is that not required/wanted here?

nizarq: Source is only a field on request telemetry (http calls). It is not a custom field. The only thing that it is used for right now is to contain the cross component correlation id (hashed ikey of the callee). Exceptions and events etc. may or may not be in the context of an http call. Also request failures are also captured by request telemetry with a failure status.

The intent so far is to capture data to drive experiences dealing with cross component communication, hence concentrating on request and dependency telemetry types. If we do in future come up with experiences where other telemetry gathered while processing the request needs such information - it can be added at that time by updating the backend schema and adding collection logic in the sdk.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
daschult: The actual call to send this request goes through [here](https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/netcore50/System/Net/HttpHandlerToFilter.cs#L52), and it doesn't look like it throws. This would tell me that OnResponseReceived() would always get called after OnRequestSent().

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
daschult: I've changed OnBeginRequest() and EndRequest() to only get their instrumentation key from the client instance. I still call "this.client.Initialize(telemetry)" in EndRequest(), just to ensure that the RequestTelemetry object has the instrumentation key set.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/InstrumentationKeyHashLookupHelper.cs
daschult: Done.

test/EmptyApp.FunctionalTests/Startup.cs
daschult: Switched to use www.example.com instead.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
daschult: I've changed this so that the telemetry is initialized just before the call to TrackRequest(). I'm not sure why the other implementations of HostingDiagnosticListener (dotnet-server) do this, but I'm just trying to follow their pattern.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
daschult: There's no instances of RequestTelemetry in the aspnet repositories that end up calling the HostingDiagnosticListener. This seems to be the only place where RequestTelemetry is created and sent to ApplicationInsights.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
daschult: TryRemove() would return false if the loggingRequestId didn't exist in the dictionary. It's highly unlikely that this will be the case (no normal code paths would create this scenario), so I don't think we need to bother writing a log for it.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
daschult: Changed.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
daschult: Added null check.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/DependencyCollectorDiagnosticListenerTest.cs
daschult: Changed to AssertTelemetryValid()

test/Microsoft.ApplicationInsights.AspNetCore.Tests/DependencyCollectorDiagnosticListenerTest.cs
daschult: Done.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/DependencyCollectorDiagnosticListenerTest.cs
daschult: Changed.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/DependencyCollectorDiagnosticListenerTest.cs
daschult: Changed.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
```diff
@@ -0,0 +1,103 @@
...
+            Uri requestUri = request.RequestUri;
+            string resourceName = requestUri.AbsolutePath;
+            if (!string.IsNullOrEmpty(httpMethod))
+            {
+                resourceName = httpMethod + " " + resourceName;
+            }
+
+            DependencyTelemetry telemetry = new DependencyTelemetry();
+            telemetry.Start();
+            this.client.Initialize(telemetry);
```
pharring: Consider using ```this.client.StartOperation<DependencyTelemetry>(resourceName)``` here. It does the ```telemetry.Start()``` call and ```this.client.Initialize```.
You get back an ```IDisposable``` operation holder which you should put in the dictionary (instead of ```DependencyTelemetry```) and ```Dispose``` in the ```OnResponseReceived``` handler.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
daschult: I can't figure out from the code or the documentation how either of these flags in the Net45 implementation would get set, so I'm just going to remove the comment for now.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/DepedencyCollectorDiagnosticListener.cs
daschult: this.client.StartOperation<DependencyTelemetry>(resourceName) isn't available in the netstandard1.6 framework, so I can't do this.

daschult: Oh, I take it back. I just found that it's an extension method. I'll look into this further.

daschult: Nope, I just looked again and it isn't available for netstandard1.6 framework. Can't do it.

pharring: It is an extension method, but you're also correct that it's not present in the netstandard version of the core Microsoft.ApplicationInsights assembly - even in version 2.3. To expose it, we would need to produce a version of the portable assembly that targets a newer netstandard (1.6 probably) and that's quite a big change.

daschult:
I've removed the DependencyCollector implementation from this pull request. I'll create a new pull request in the [ApplicationInsights-dotnet-server](https://github.com/Microsoft/ApplicationInsights-dotnet-server) repository when its ready.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -38,10 +39,19 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
...
             {
-                httpContext.Features.Set(new RequestTelemetry());
+                RequestTelemetry requestTelemetry = new RequestTelemetry();
+                this.client.Initialize(requestTelemetry);
+                requestTelemetry.Id = httpContext.TraceIdentifier;
+                httpContext.Features.Set(requestTelemetry);
 
                 this.beginRequestTimestamp.Value = timestamp;
-                this.client.Context.Operation.Id = httpContext.TraceIdentifier;
+                this.client.Context.Operation.Id = requestTelemetry.Id;
```
pharring: I just want to make you aware of #343 which is waiting to merge and also makes changes in this block of code. If #343 is merged first you'll almost certainly have a conflict. I can help with the final merge.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -37,14 +38,19 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
         {
             if (this.client.IsEnabled())
             {
-                var requestTelemetry = new RequestTelemetry
+                var requestTelemetry = new RequestTelemetry()
```
pharring: Nit: The parentheses are optional when using initializer syntax.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -37,14 +38,19 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
...
                 {
                     Id = httpContext.TraceIdentifier
                 };
-
                 requestTelemetry.Start(/*timestamp*/);
-
                 httpContext.Features.Set(requestTelemetry);
+
+                this.client.Initialize(requestTelemetry);
```
pharring: Could you put the call to ```Intialize``` before the call to ```Start```? This is because of https://github.com/Microsoft/ApplicationInsights-dotnet/pull/431 (waiting to be merged). ```Start``` now sends ETW events and the payload includes Operation ID and Parent Operation ID which may be set in initializers.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
daschult: Removed.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
daschult: Done.

daschult:
@pakrym @brahmnes @dnduffy I've removed the DependencyCollector implementation and responded to each of the comments. Please take another look at this pull request and merge if it's satisfactory.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -41,10 +42,15 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
...
                     Id = httpContext.TraceIdentifier
                 };
-
+                this.client.Initialize(requestTelemetry);
                 requestTelemetry.Start(/*timestamp*/);
-
                 httpContext.Features.Set(requestTelemetry);
+
+                IHeaderDictionary responseHeaders = httpContext.Response?.Headers;
+                if (responseHeaders != null && !responseHeaders.ContainsKey(RequestResponseHeaders.TargetInstrumentationKeyHeader) && !string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey))
```
pharring: Nit: For perf, put the ```IsNullOrEmpty``` check before the ```ContainsKey``` check. C# has guaranteed left-to-right evaluation and short-circuiting.

daschult: Done.

pharring:
LGTM 👍

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -109,6 +115,18 @@ private void EndRequest(HttpContext httpContext, long timestamp)
                     telemetry.Success &= successExitCode;
                 }
 
+                IHeaderDictionary requestHeaders = httpContext.Request?.Headers;
```
SergeyKanzhelev: Can you please follow this PR: https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/326 to allow customer to override the `Source`

SergeyKanzhelev: Some clarifications: we changed the implementation of this logic in classic ASP.NET to allow customer to override this value. As minimum - check that this value is not set yet by telemetry initializers. Ideally - move this logic to `OperationCorrelationTelemetryInitializer`

SergeyKanzhelev: @cithomas can provide more details

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/OperationCorrelationTelemetryInitializer.cs
```diff
@@ -49,6 +50,20 @@ public OperationCorrelationTelemetryInitializer(IHttpContextAccessor httpContext
                     telemetry.Context.Operation.Id = parentContext.Id;
                 }
             }
+
+            HttpRequest currentRequest = platformContext.Request;
```
SergeyKanzhelev: Please create/update unit tests for this telemetry initializer?

PR #343, opened 2017-02-05
Allow simultaneous overlapping requests in HostingDiagnosticListener

### LINKED ISSUE #333: Telemetry does not include "ai.operation.parentId".
Telemetry items sent from the .NET Core SDK do not have the "ai.operation.parentId" tag.

### LINKED ISSUE #340: Request-Dependency correlation is broken
Operation id of the request is not propagated to the child dependencies. Please find the following example:
image

### LINKED ISSUE #342: Operation IDs are being duplicated under load
I notice that, under load, my ASP.NET Core website generates request telemetry with duplicate Operation IDs.

It appears that HostingDiagnosticListener isn't written with multiple simultaneous requests in mind.
In particular, ```this.client.Context.Operation.Id``` gets overwritten by each new request ([line 44](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/master/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs#L44))
(Also ```beginRequestTimestamp```)

Rather that setting the Operation ID in the client's context (which is like setting it globally), I believe it should be set in the ```RequestTelemetry``` object that is added to ```HttpContext.Features```
One would think that the timestamp could also be recorded in the RequestTelemetry object (indeed it already has fields for recording start time, so it's odd that it's not initialized here)

pharring:
Fixes #342 
Fixes #333 
Fixes #340


src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -93,8 +97,8 @@ private void EndRequest(HttpContext httpContext, long timestamp)
                     return;
                 }
 
-                telemetry.Duration = new TimeSpan(timestamp - this.beginRequestTimestamp.Value);
-                telemetry.Timestamp = DateTime.Now - telemetry.Duration;
+                telemetry.Timestamp = telemetry.StartTime;
```
SergeyKanzhelev: `StartTime` is deprecated. In this line you assigning the value to itself

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -13,8 +13,12 @@
     /// </summary>
     internal class HostingDiagnosticListener : IApplicationInsightDiagnosticListener
     {
+        /// <summary>
+        /// Conversion from stopwatch time (ticks since boot) to local time.
+        /// </summary>
+        private static readonly DateTimeOffset stopwatchEpoch = DateTime.Now - TimeSpan.FromTicks(Stopwatch.GetTimestamp());
```
SergeyKanzhelev: We found issues with this approach - it will work badly when windows adjusts the clock (some machines has bad bios and time constantly running behind or ahead) or when there is a summer time switch. Mostly the first though.

You can use extension methods `RequestTelemetry.Start()` and `Stop()` instead. See https://github.com/Microsoft/ApplicationInsights-dotnet/blob/dd654d2c0cd47623d0e99e6e19481cba3203fa98/src/Core/Managed/Shared/OperationTelemetryExtensions.cs#L19

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -38,10 +42,10 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
         {
             if (this.client.IsEnabled())
             {
-                httpContext.Features.Set(new RequestTelemetry());
-
-                this.beginRequestTimestamp.Value = timestamp;
-                this.client.Context.Operation.Id = httpContext.TraceIdentifier;
+                var requestTelemetry = new RequestTelemetry();
+                requestTelemetry.Context.Operation.Id = httpContext.TraceIdentifier;
```
SergeyKanzhelev: It should be `requestTelemetry.Id = httpContext.TraceIdentifier;` This will fix https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/340 Also consider fixing https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/333 in this PR - I'd appreciate it!

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
pharring: OK, will do. Note that using the Start/Stop extensions, means that we'll ignore the timestamp being passed into OnBeginRequest/OnEndRequest.

SergeyKanzhelev: I didn't realize this. @pakrym - what was the reasoning in passing that timestamp? Does it represent more precise time of the request beginning?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
pharring: OK, but I feel obliged to point out that ```OperationTelemetryExtensions.Start()``` sets both :-)

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
pharring: I can do that, but I'm unclear what the rules should be here. I had assumed that the original code was correct in setting the operation Id (except that it wasn't correct for simultaneous requests). Should we set both? Or was the original just completely wrong?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
pakrym: Timestamp is more precise and is faster then `DateTime.Now`

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
pharring: Answering my own question, it looks like the original was incorrect. We have an ```OperationIdTelemetryInitializer``` which will supply an OperationId. So, #340 is fixed (I'll also add a test for it).

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
pharring: Yes, the timestamp comes from: 
https://github.com/aspnet/Hosting/blob/4abb48e1aaa730062cef7f0f0fd08c9c5b72413d/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs#L37

For now, I'll leave it the way it is (although I will change it to use UTC to avoid time-zone ambiguity). I see the changes made to support the higher resolution clock in the shared library: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/358 

Internally, there's a ```BeginTimeInTicks``` property on ```OperationTelemetry``` that ```Start()``` and ```Stop()``` set. It would be nice if those extension methods had overloads that would let me supply the timestamp in ticks, thus avoiding a second call to ```Stopwatch.GetTimestamp()```

ledouxpl: Please use Start/Stop.

I was planning to go in production with v2.0.0 and realized the bug with the RequestTelemetry and DependencyTelemetries not correlating the OperationId correctly and an issue with the RequestTelemetries timestamp being set after the one set on the related DependencyTelemetry instances.

I was about to implement my own fix but noticed your pull request which brings in the OperationCorrelationTelemetryInitializer. I also noticed the changes you made to the HostingDiagnosticListener.
I cloned your pull request branch in hope of it fixing the timing issues, but it is worse (at least when testing locally) since Stopwatch.GetTimestamp() returns the timestamp of my PC’s last boot.

Using the RequestTelemetry Start/Stop extension instead of the timespan rectified the situation on my end.
I understand wanting to use the OnBeginRequest and OnEndRequest timestamp parameters, but this just not seem to be working on my end. I am without doubt missing key elements here to fully grasp the full impact of this choice, but I fail to see how using Stopwatch.GetTimestamp() will give consistent, testable results.

Since IApplicationInsightDiagnosticListener is internal, I cannot simply remove from the services collection your implementation and add in my own using Start/Stop.

pharring: OK, I have switched to using Start/Stop. I had to disable one of the tests until the overloads I added in https://github.com/Microsoft/ApplicationInsights-dotnet/pull/424 are available in a nuget we can consume from here. In testing, I also discovered that we had a precision problem in Stop() (it was rounding durations to the nearest millisecond) which I fixed with https://github.com/Microsoft/ApplicationInsights-dotnet/pull/426.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
pharring: Ah... it used to in an older SDK. In 2.2 it sets only Timestamp.

pharring:
@SergeyKanzhelev can you review again, please?

pharring:
@SergeyKanzhelev do you approve the changes? It works as written, but it's not using the timestamp from the host. The timestamp overloads of Start/Stop are not in 2.3.0-Beta2, so I can't use them (yet). When they're available, it's a simple matter of uncommenting their usage in HostingDiagnosticsListener and re-enabling the two skipped unit tests.

pharring:
@dnduffy @SergeyKanzhelev Merge please?

PR #376, opened 2017-03-30
Use app id than hashed iKey for correlationId

xiaomi7732:
This will work with the dependency collector here:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/366
for asp.net core application to support multiple server nodes in App-Map when servers instrumented with different iKeys.


xiaomi7732:
Please do NOT merge this PR. 
1. This will need to wait until Microsoft/ApplicationInsights-dotnet-server#366 is in.
2. Unit tests are not updated.

I send this PR out mainly seeking for early feedback for the code implementation. Please comment on the implementations for the code.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,199 @@
...
+    using System.Globalization;
+    using System.Threading.Tasks;
+    
+    using Microsoft.ApplicationInsights.AspNetCore.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Extensibility;
+
+#if NET451
+    using System.IO;
+    using System.Net;
+#else
```
pharring: Is this really applicable to .net core?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// </summary>
+        /// <param name="instrumentationKey">Instrumentation key for which app id is to be retrieved.</param>
+        /// <returns>App id.</returns>
+        private async Task<string> FetchAppIdFromService(string instrumentationKey)
+        {
+            try
+            {
+                Uri appIdEndpoint = this.GetAppIdEndPointUri(instrumentationKey);
+                string result = null;
+#if NET451
```
pharring: Is this appropriate for .Net core?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,199 @@
...
+                try
+                {
+
+                    // We wait for <getAppIdTimeout> seconds (which is 0 at this point) to retrieve the appId. If retrieved during that time, we return success setting the correlation id.
+                    // If we are still waiting on the result beyond the timeout - for this particular call we return the failure but queue a task continuation for it to be cached for next time.
+                    Task<string> getAppIdTask = this.provideAppId(instrumentationKey.ToLowerInvariant());
+                    if (getAppIdTask.Wait(GetAppIdTimeout))
+                    {
+                        this.GenerateCorrelationIdAndAddToDictionary(instrumentationKey, getAppIdTask.Result);
+                        correlationId = this.knownCorrelationIds[instrumentationKey];
```
pharring: Suggestion: Modify ```GenerateCorrelationIdAndAddToDictionary``` to return the generated correlationId so you don't have to look it up in the dictionary again.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs
```diff
@@ -0,0 +1,199 @@
...
+                {
+                    this.knownCorrelationIds.Clear();
+                }
+
+                try
+                {
+
+                    // We wait for <getAppIdTimeout> seconds (which is 0 at this point) to retrieve the appId. If retrieved during that time, we return success setting the correlation id.
+                    // If we are still waiting on the result beyond the timeout - for this particular call we return the failure but queue a task continuation for it to be cached for next time.
+                    Task<string> getAppIdTask = this.provideAppId(instrumentationKey.ToLowerInvariant());
```
pharring: What's to stop duplicate tasks being created here? i.e. You call ```TryGetXComponentCorrelationId``` for a new instrumentationKey and a task is created. Before it completes, you call it again for the same instrumentation key. Idea: Add a "pending" value in the dictionary to indicate that a task is in flight. I might even go as far as making it a ```ConcurrentDictionary<string, Task<string>>``` although I'm not sure about the memory overhead. (Tasks can hold onto a lot of state).

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HeadersUtilities.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <returns>The result of setting the provided key name/value pair into the provided headerValues.</returns>
+        public static IEnumerable<string> SetHeaderKeyValue(IEnumerable<string> headerValues, string keyName, string keyValue)
+        {
+            string[] newHeaderKeyValue = new[] { string.Format(CultureInfo.InvariantCulture, "{0}={1}", keyName.Trim(), keyValue.Trim()) };
+            return headerValues == null || !headerValues.Any()
+                ? newHeaderKeyValue
+                : headerValues
+                    .Where((string headerValue) =>
+                    {
+                        int equalsSignIndex = headerValue.IndexOf('=');
```
pharring: Always use the overload of ```IndexOf``` that takes a ```StringComparison```. In this case ```StringComparison.Ordinal``` would be appropriate. You avoid a more expensive locale-sensitive search.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HttpHeadersUtilities.cs
```diff
@@ -0,0 +1,112 @@
+namespace Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+#if NET451
```
pharring: Delete?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HttpHeadersUtilities.cs
```diff
@@ -0,0 +1,112 @@
...
+#endif
+    using System.Net.Http.Headers;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.Extensions.Primitives;
+#if NETCORE
```
pharring: Is this really necessary? I thought this was all .Net Core.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HttpHeadersUtilities.cs
```diff
@@ -0,0 +1,112 @@
...
+    using Microsoft.ApplicationInsights.Common;
+#endif
+    internal static class HttpHeadersUtilities
+    {
+        internal static IEnumerable<string> GetHeaderValues(HttpHeaders headers, string headerName)
```
pharring: Consider making these methods extension methods.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HeadersUtilities.cs
```diff
@@ -0,0 +1,87 @@
...
+                        return equalsSignIndex == -1 || TrimSubstring(headerValue, 0, equalsSignIndex) != keyName;
```
pharring: Avoid creating temporary strings just to do an equality check. Instead, do a custom comparison on the original string. (It's the same thing as doing ```s1.Equals(s2, StringComparison.InvariantCultureIgnoreCase)``` instead of ```s1.ToLowerInvariant().Equals(s2.ToLowerInvariant())```.)

pharring: Try this:
```C#
bool HeaderMatchesKey(string header, string key)
{
    int equalsSignIndex = header.IndexOf('=');
    if (equalsSignIndex < 0)
    {
        return false;
    }
 
    // Skip leading whitespace
    int start;
    for (start = 0; start < equalsSignIndex; start++)
    {
        if (!char.IsWhiteSpace(header[start]))
        {
            break;
        }
    }

    if (string.CompareOrdinal(header, start, key, 0, key.Length) != 0)
    {
        return false;
    }

    // Check trailing whitespace
    for (int i = start + key.Length; i < equalsSignIndex; i++)
    {
        if (!char.IsWhiteSpace(header[i]))
        {
            return false;
        }
    }

    return true;
}

File: src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs (hunk 1 above)
Comment (xiaomi7732) [inline src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs]: The codebase, in fact, supports both net451 & netstandard1.6 as in project.json:
-
    "frameworks": {
        "net451": {
            "dependencies": {
                "Microsoft.ApplicationInsights.DependencyCollector": "2.3.0-beta3",
                "Microsoft.ApplicationInsights.PerfCounterCollector": "2.3.0-beta3",
                "Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel": "2.3.0-beta3"
            }
        },
        "netstandard1.6": {
            "dependencies": {
                "NETStandard.Library": "1.6.0",
                "System.Net.NameResolution": "4.0.0"
            }
        }
    }
-
Since this is a general helper method, I think we just keep it working for both target frameworks.

File: src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs (hunk 2 above)
Comment (xiaomi7732) [inline src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs]: Same reason.

File: src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs (hunk 3 above)
Comment (xiaomi7732) [inline src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs]: Good idea. Suggestion accepted. 💯

File: src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs (hunk 4 above)
Comment (xiaomi7732) [inline src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/CorrelationIdLookupHelper.cs]: I see the value here to prevent task querying the same app id being scheduled. Code updated in the next iteration.
Thanks!

File: src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HttpHeadersUtilities.cs (hunk 7 above)
Comment (xiaomi7732) [inline src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HttpHeadersUtilities.cs]: Same reason. Refer project.json.

File: src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HttpHeadersUtilities.cs (hunk 8 above)
Comment (xiaomi7732) [inline src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HttpHeadersUtilities.cs]: Thanks for the point. For now, I don't see big advantage of extension method and I don't see existing pattern for writing utility as extension methods. Won't fix.
Let me know if you have a strong preference on that.

File: src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HttpHeadersUtilities.cs (hunk 6 above)
Comment (xiaomi7732) [inline src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HttpHeadersUtilities.cs]: Oach! Thanks for the catch.

File: src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HeadersUtilities.cs (hunk 9 above)
Comment (xiaomi7732) [inline src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HeadersUtilities.cs]: Good point. For the sake of the better performance, I have stolen your code here, haha. Change will show up in the next iteration.

**Comment by xiaomi7732** [comment]
Code issues in review addressed. Unit tests repaired and 2 added. Please review the code and comment if there's any concern. Otherwise, please kindly merge this PR. Thanks!

File: src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs [hunk 10]
```diff
@@ -45,11 +48,16 @@ public void OnBeginRequest(HttpContext httpContext, long timestamp)
                 this.client.Initialize(requestTelemetry);
                 requestTelemetry.Start(timestamp);
                 httpContext.Features.Set(requestTelemetry);
-
                 IHeaderDictionary responseHeaders = httpContext.Response?.Headers;
-                if (responseHeaders != null && !string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey) && !responseHeaders.ContainsKey(RequestResponseHeaders.TargetInstrumentationKeyHeader))
+                if (responseHeaders != null &&
+                    !string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey) &&
+                    (!responseHeaders.ContainsKey(RequestResponseHeaders.RequestContextHeader) || HttpHeadersUtilities.ContainsRequestContextKeyValue(responseHeaders, RequestResponseHeaders.RequestContextTargetKey)))
```
dnduffy: So the helper method does an iteration through all of the values, did the old way do a hash lookup?  Which is less expensive?

xiaomi7732: The current implementation is more expensive in the sense of dealing different types of headers. As before, the header's we are interested is a dictionary and now it's single header contains sets of key value pairs.

|           | Key              	| Value                        	
|------- |------------------	|------------------------------	
|before | x-request-source 	| http://www.source.com        	
|after| context-header 	| source=http://www.source.com, blala=balabala 	

And that is why we build the utility to help us getting the value out.
With regard why we change the header format, I think that is off the topic of this specific change.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/OperationCorrelationTelemetryInitializer.cs
```diff
@@ -1,22 +1,34 @@
...
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
     using Microsoft.AspNetCore.Http;
+    using Microsoft.Extensions.Primitives;
 
     /// <summary>
     /// A telemetry initializer that will set the correlation context for all telemetry items in web application.
     /// </summary>
     internal class OperationCorrelationTelemetryInitializer : TelemetryInitializerBase
```
dnduffy: This is the higher level abstraction we should target for sharing at, we should only have one copy of this class and reuse it from the other package.

PR #377, opened 2017-03-31
Merge develop to master for 2.1.0 release.

dnduffy:


test/MVCFramework45.FunctionalTests/Views/Home/Dependency.cshtml
```diff
@@ -0,0 +1,2 @@
+﻿@{
```
SergeyKanzhelev: why this file is empty? do we need it?

dnduffy: I believe unit tests are calling it to record a dependency and it didn't need any special content of its own (it would get the layout from _layout.cshtml) for the purpose of the test call.

PR #385, opened 2017-04-14
Initialize RequestTelemetry with CorrelationId information

yantang-msft:
@dnduffy, Here is the change we made that upgraded the reference of AspNetCore to 2.0


src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -1,108 +1,108 @@
...
-        "netstandard1.6": {
-            "dependencies": {
-                "NETStandard.Library": "1.6.0",
-                "System.Net.NameResolution": "4.0.0"
-            }
-        }
+  "frameworks": {
+    "net46": {
+      "dependencies": {
+        "Microsoft.ApplicationInsights.DependencyCollector": "2.4.0-beta1-build09989",
```
yantang-msft: The DependencyCollector will not work if we target "net46", because it pick up the dll from the Net45 folder, which is not compatible with the System.Net.Http that is brought by AspNetCore 2.0

test/FunctionalTestUtils/TelemetryTestsBase.cs
```diff
@@ -23,6 +23,28 @@ public abstract class TelemetryTestsBase
...
+        /// <summary>
+        /// Ideally, when the WebHost sever get disposed, all requests will be finished.
+        /// However, the current behavior is server dispose doesn't wait for the requests to be finished.
+        /// And during server disposing, the injected AI service will also get disposed, which means some telemetry might not be sent after server get disposed.
+        /// So we add this helper funciton to make sure all telemtry are sent before checking anything.
+        /// </summary>
+        /// <param name="condition">The condition to meet</param>
+        /// <param name="timeoutMs">Timeout in ms</param>
+        /// <param name="intervalMs">Sleep interval in ms</param>
+        public static void ConditionalTimeout(Func<bool> condition, int timeoutMs = 10000, int intervalMs = 50)
```
pharring: Isn't this equivalent to ```SpinWait.SpinUntil(..)``` ?

test/FunctionalTestUtils/TelemetryTestsBase.cs
```diff
@@ -23,6 +23,28 @@ public abstract class TelemetryTestsBase
...
+        {
+            DateTime timeout = DateTime.Now.AddMilliseconds(timeoutMs);
+            while (DateTime.Now < timeout)
```
pharring: ```Never use DateTime.Now``` for timing loops. It's slow, it has discontinuities and can be adjusted at any point.
```DateTime.UtcNow``` is better but, for this, you should be using ```System.Diagnostics.Stopwatch```.

File: test/FunctionalTestUtils/TelemetryTestsBase.cs [hunk 4]
```diff
@@ -23,6 +23,28 @@ public abstract class TelemetryTestsBase
...
+            {
+                if (condition())
+                {
+                    return;
+                }
+                Thread.Sleep(intervalMs);
+            }
```
pharring: Suggest: ```throw new TimeoutException();```
Or, have this method return a ```bool``` (true=>condition met; false=>timeout) and ```Assert.IsTrue``` in the caller.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -34,19 +36,66 @@ public HostingDiagnosticListener(TelemetryClient client, ICorrelationIdLookupHel
...
+                        requestTelemetry.Context.Operation.ParentId = xmsRequestId.Value[0];
+                    }
+                }
+
+                requestTelemetry.Id = Activity.Current?.Id;
+                requestTelemetry.Context.Operation.Id = Activity.Current?.RootId;
+
                 this.client.Initialize(requestTelemetry);
-                requestTelemetry.Start(timestamp);
+                requestTelemetry.Start(Stopwatch.GetTimestamp());
```
pharring: It's a shame that we've lost the 'external' timestamp on this method because it allowed us to get very precise timings. I presume it was an intentional change on the ASP.NET Core side (the originator of this HttpRequestIn.Start message).

lmolkova: It was intentionally lost  since Activity exposes StartTimeUtc property.
However there are certain issues with it: 
UtcNow precision was improved: https://github.com/dotnet/coreclr/pull/9736 (previously is was ~16ms)

However if AI SDK runs on .NET Desktop, it's not going to get precise timestamp. We'll rely on the local timstamp for now.

yantang-msft: They don't pass the timestamp because the created Activity has StartTimeUtcNow property. But you're right, it feels less accurate. I can create an issue for AspNetCore if you think it's important.

test/FunctionalTestUtils/TelemetryTestsBase.cs
yantang-msft: It's the same... So much better to use this function instead.

test/FunctionalTestUtils/TelemetryTestsBase.cs
yantang-msft: Changed to SpinUntil

test/FunctionalTestUtils/TelemetryTestsBase.cs
yantang-msft: Good suggestion! 👍

PR #386, opened 2017-04-14
Added null checks for incorrectly initialized AI ILogger.

dnduffy:


test/Microsoft.ApplicationInsights.AspNetCore.Tests/Logging/ApplicationInsightsLoggerTests.cs
```diff
@@ -27,5 +27,15 @@ public void TestLoggerSetsSdkVersionOnLoggedTelemetry()
...
             Assert.True(isCorrectVersion);
         }
+
+        /// <summary>
+        /// Tests that an incorrectly constructed or uninitialized Application Insights ILogger does not throw exceptions.
+        /// </summary>
+        [Fact]
+        public void TestUninitializedLoggerDoesNotThrowExceptions()
+        {
+            ILogger logger = new ApplicationInsightsLogger("test", null, null);
```
Dmitry-Matveev: Do we need to test out both telemetry client and filter being null separately?

PR #400, opened 2017-04-20
Initialize RequestTelemetry with correlation information

yantang-msft:
Initialize RequestTelemetry's correlation information based on the new Request-Id header.
If the running AspNetCore is 1.0, only the old event "Microsoft.AspNetCore.Hosting.BeginRequest" will be sent. We will create an activity which will be used by the dependency collector.
If the running AspNetCore is 2.0, both the old event and the new event "Microsoft.AspNetCore.Hosting.HttpRequestIn.Start" will be sent. In this case, we ignore the old event so we don't duplicate the RequestTelemetry. AspNetCore will also create an activity for us in this case.
We will also look for standard headers in case the calling service doesn't have the latest AI sdk.


src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -128,7 +128,6 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
             services.AddSingleton<ITelemetryInitializer, DomainNameRoleInstanceTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, ComponentVersionTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, ClientIpHeaderTelemetryInitializer>();
-            services.AddSingleton<ITelemetryInitializer, OperationIdTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, OperationNameTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, ApplicationInsights.AspNetCore.TelemetryInitializers.OperationCorrelationTelemetryInitializer>();
```
lmolkova: Please check OperationCorrelationTelemetryInitializer and remove operation Id, root id and parent Id initialization from there

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -76,32 +76,33 @@
...
+                "Microsoft.ApplicationInsights.DependencyCollector": "2.4.0-beta1",
+                "Microsoft.ApplicationInsights.PerfCounterCollector": "2.4.0-beta1",
+                "Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel": "2.4.0-beta1"
             }
         },
         "netstandard1.6": {
             "dependencies": {
-                "NETStandard.Library": "1.6.0",
-                "System.Net.NameResolution": "4.0.0"
+                "NETStandard.Library": "1.6.1",
```
lmolkova: reference DependencyCollector here and make sure tests for DependenycCollector on netstandard1.6 are enabled

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -17,6 +21,8 @@ internal class HostingDiagnosticListener : IApplicationInsightDiagnosticListener
         private readonly TelemetryClient client;
         private readonly ICorrelationIdLookupHelper correlationIdLookupHelper;
         private readonly string sdkVersion;
+        private bool? isAspNetCore20;
```
lmolkova: could you make it static and initialize in place, so it don't have to be nullable, you can also remove property

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -34,31 +40,67 @@ public HostingDiagnosticListener(TelemetryClient client, ICorrelationIdLookupHel
...
+        {
+            get
+            {
+                if (isAspNetCore20.HasValue)
+                {
+                    return isAspNetCore20.Value;
+                }
+
+                var hostingVersion = typeof(WebHostBuilder).GetTypeInfo().Assembly.GetName().Version;
+                isAspNetCore20 = hostingVersion.Major >= 2;
```
lmolkova: 👍

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -99,6 +150,124 @@ public void OnDiagnosticsUnhandledException(HttpContext httpContext, Exception e
             this.OnException(httpContext, exception);
         }
 
+        private RequestTelemetry InitializeRequestTelemetryOnHttpRequestInStart(HttpContext httpContext)
+        {
+            var requestTelemetry = new RequestTelemetry();
+
+            if (Activity.Current == null)
```
lmolkova: store Activity.Current in local variable: it's AsyncLocal and accessing nultiple times it is not completely cheap

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -99,6 +150,124 @@ public void OnDiagnosticsUnhandledException(HttpContext httpContext, Exception e
...
+                }
+
+                StringValues xmsRequestId;
+                if (httpContext.Request.Headers.TryGetValue(RequestResponseHeaders.StandardParentIdHeader, out xmsRequestId))
+                {
+                    requestTelemetry.Context.Operation.ParentId = xmsRequestId;
+                }
+            }
+
+            requestTelemetry.Id = Activity.Current?.Id;
```
lmolkova: Activity.Current is never null at this point

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -99,6 +150,124 @@ public void OnDiagnosticsUnhandledException(HttpContext httpContext, Exception e
             this.OnException(httpContext, exception);
         }
 
+        private RequestTelemetry InitializeRequestTelemetryOnHttpRequestInStart(HttpContext httpContext)
+        {
+            var requestTelemetry = new RequestTelemetry();
+
+            if (Activity.Current == null)
+            {
+                AspNetCoreEventSource.Instance.LogHostingDiagnosticListenerOnHttpRequestInStartActivityNull();
```
lmolkova: this means something really strange happened and AspNEtCore did not create activity for us but sent the event. return, there is nothing else we can do

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
lmolkova: must not be null as you return before

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -99,6 +150,124 @@ public void OnDiagnosticsUnhandledException(HttpContext httpContext, Exception e
...
+                            activity.AddBaggage(baggageItem.Name, baggageItem.Value);
+                            if (!requestTelemetry.Context.Properties.ContainsKey(baggageItem.Name))
+                            {
+                                requestTelemetry.Context.Properties[baggageItem.Name] = baggageItem.Value;
+                            }
+                        }
+                    }
+                }
+            }
+            else if (httpContext.Request.Headers.TryGetValue(RequestResponseHeaders.StandardRootIdHeader, out standardRootId))
```
lmolkova: 1. I think it's almost exactly the same code as on line 179 in `InitializeRequestTelemetryOnHttpRequestInStart`. Could you have a separate function that parses legacy headers

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -34,31 +40,67 @@ public HostingDiagnosticListener(TelemetryClient client, ICorrelationIdLookupHel
...
+
+        /// <summary>
+        /// Diagnostic event handler method for 'Microsoft.AspNetCore.Hosting.HttpRequestIn.Start' event.
+        /// </summary>
+        [DiagnosticName("Microsoft.AspNetCore.Hosting.HttpRequestIn.Start")]
+        public void OnHttpRequestInStart(HttpContext httpContext)
+        {
+            if (this.client.IsEnabled())
+            {
+                var requestTelemetry = InitializeRequestTelemetryOnHttpRequestInStart(httpContext);
```
lmolkova: is there any point in having additional `InitializeRequestTelemetryOnHttpRequestInStart` method? 
It is used just once and makes it harder to read code. 

The same applies to `InitializeRequestTelemetryOnBeginRequest`

However if you prefer to keep it like this, I'm also fine with it.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -99,6 +150,124 @@ public void OnDiagnosticsUnhandledException(HttpContext httpContext, Exception e
...
+        private void SetAppIdInResponseHeader(HttpContext httpContext, RequestTelemetry requestTelemetry)
+        {
+            IHeaderDictionary responseHeaders = httpContext.Response?.Headers;
+            if (responseHeaders != null &&
+                !string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey) &&
+                (!responseHeaders.ContainsKey(RequestResponseHeaders.RequestContextHeader) || HttpHeadersUtilities.ContainsRequestContextKeyValue(responseHeaders, RequestResponseHeaders.RequestContextTargetKey)))
+            {
+                string correlationId = null;
+                if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(requestTelemetry.Context.InstrumentationKey, out correlationId))
+                {
```
lmolkova: OperationCorrelationTelemetryInitializer also queries correlatioIdLookupHelper for the same instrumentation key.
I wonder if code from OperationCorrelationTelemetryInitializer could go here and then we can just get rid of it?

src/Microsoft.ApplicationInsights.AspNetCore/project.json
yantang-msft: Will do this in another PR, I have seen issues with unit tests if doing so.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
yantang-msft: Done

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: Sure

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: good to know

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: I'll just remove this. It shouldn't happen

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: Yep

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: It's not the same. In new events code path, we create another activity. In old events code path, only one Activity is created for both old and new headers.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: I feel this part is a bit long. And I want to give it a function name so people know what it is doing.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: OperationCorrelationTelemetryInitializer set Request.Source, here it sets Request.Reponse.Header. But basically, they are doing similar things. I guess this part of code can be moved into OperationCorrelationTelemetryInitializer, and I'm willing to remove the additional "InitializeRequestTelemetryOnBeginRequest" in this case.
However, the logic here will override the value, not like other initializers which first checks if the value is null or empty. So I don't want to take the risk of moving it to the initializer.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -78,7 +123,13 @@ public void OnEndRequest(HttpContext httpContext, long timestamp)
         public void OnHostingException(HttpContext httpContext, Exception exception)
         {
             this.OnException(httpContext, exception);
-            this.EndRequest(httpContext, Stopwatch.GetTimestamp());
+
+            // In AspNetCore 1.0, when an exception is unhandled, it will only send UnhandledException event but not EndRequest event, so we need to EndRequest at here.
+            // In AspNetCore 2.0, after sending UnhandledException, it will stop the created activity, which will send HttpRequestIn.Stop event, so we just EndRequest at there.
```
dnduffy: nitpick: "...send the HttpRequestIn.Stop event, so we will just end the request there."

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -77,6 +62,12 @@ public void OnHttpRequestInStart(HttpContext httpContext)
         {
             if (this.client.IsEnabled())
             {
+                if (Activity.Current == null)
+                {
+                    AspNetCoreEventSource.Instance.LogHostingDiagnosticListenerOnHttpRequestInStartActivityNull();
+                    return;
+                }
+
                 var requestTelemetry = InitializeRequestTelemetryOnHttpRequestInStart(httpContext);
```
lmolkova: I still believe  InitializeRequestTelemetryOnHttpRequestInStart and InitializeRequestTelemetryOnBeginRequest are almost the same:
Basically InitializeRequestTelemetryOnBeginRequest  is 
```
InitializeRequestTelemetryOnBeginRequest (context, timestamp)
{
  var a = new Activity(...);
  // get Request-Id and Correlation-Context headers
  a.Start();
  InitializeRequestTelemetryOnHttpRequestInStart (context, timestamp)
}
```

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: Modified :)

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: It's somehow consolidated. :)

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -14,9 +18,17 @@
     /// </summary>
     internal class HostingDiagnosticListener : IApplicationInsightDiagnosticListener
     {
+        /// <summary>
+        /// Determine whether the running AspNetCore Hosting version is 2.0 or higher. This will affect what DiagnosticSource events we receive.
+        /// To support AspNetCore 1.0 and 2.0, we listen to both old and new events.
+        /// If the running AspNetCore version is 2.0, both old and new events will be sent. In this case, we will ignore the old events.
+        /// </summary>
+        public static bool IsAspNetCore20 = typeof(WebHostBuilder).GetTypeInfo().Assembly.GetName().Version.Major >= 2;
```
Dmitry-Matveev: Should we stay more conservative and keep it "==2" in case of the breaking changes and new events in 3.0 that are not accounted for in our logic below?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -34,31 +46,99 @@ public HostingDiagnosticListener(TelemetryClient client, ICorrelationIdLookupHel
...
-                IHeaderDictionary responseHeaders = httpContext.Response?.Headers;
-                if (responseHeaders != null &&
-                    !string.IsNullOrEmpty(requestTelemetry.Context.InstrumentationKey) &&
-                    (!responseHeaders.ContainsKey(RequestResponseHeaders.RequestContextHeader) || HttpHeadersUtilities.ContainsRequestContextKeyValue(responseHeaders, RequestResponseHeaders.RequestContextTargetKey)))
+                var activity = new Activity(ActivityCreatedByHostingDiagnosticListener);
+                var isActivityCreatedFromRequestIdHeader = false;
+
+                StringValues requestId;
+                StringValues standardRootId;
+                if (httpContext.Request.Headers.TryGetValue(RequestResponseHeaders.RequestIdHeader, out requestId))
```
Dmitry-Matveev: Is not-null `httpContext` a guarantee of `BeginRequest` event in diagnostic source?

yantang-msft: The instrumentation code will always send the httpContext, but I'm not sure whether it could be null at there. Since there is never a null check for httpContext, I think it's almost guaranteed.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: It's more likely AspNetCore will preserve the new events, so it's better to fall back to the new events.
Anyways, this should be removed when we upgrade the AspNetCore dependency to > 2.0 and only listening to new events. And if we don't restrict user using AspNetCore 3.0 with this library, it's equivalent to ">= 2" here.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
dnduffy: The framework guarantees it.  No null check needed.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -34,31 +46,99 @@ public HostingDiagnosticListener(TelemetryClient client, ICorrelationIdLookupHel
...
                 {
-                    string correlationId = null;
-                    if (this.correlationIdLookupHelper.TryGetXComponentCorrelationId(requestTelemetry.Context.InstrumentationKey, out correlationId))
+                    isActivityCreatedFromRequestIdHeader = true;
+                    activity.SetParentId(requestId);
+
+                    string[] baggage = httpContext.Request.Headers.GetCommaSeparatedValues(RequestResponseHeaders.CorrelationContextHeader);
```
dnduffy: nitpick: I would probably have cached httpContext.Request.Headers in a variable instead of dereferencing it each time.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -78,7 +161,13 @@ public void OnEndRequest(HttpContext httpContext, long timestamp)
         public void OnHostingException(HttpContext httpContext, Exception exception)
         {
             this.OnException(httpContext, exception);
-            this.EndRequest(httpContext, Stopwatch.GetTimestamp());
+
+            // In AspNetCore 1.0, when an exception is unhandled, it will only send UnhandledException event but not EndRequest event, so we need to EndRequest at here.
```
dnduffy: nitpick: "...when an exception is unhandled it will only send the UnhandledException event, but not the EndRequest event, so we need to call EndRequest here."

src/Microsoft.ApplicationInsights.AspNetCore/project.json
```diff
@@ -76,18 +76,18 @@
...
+        "type": "build"
+      }
+      //"Text.Analyzers": {
+      //    "version": "1.2.0-beta2",
+      //    "type": "build"
+      //},
+      //"XmlDocumentationComments.Analyzers": {
+      //    "version": "1.2.0-beta2",
+      //    "type": "build"
+      //},
```
dnduffy: nitpick: this formatting didn't need to change, especially not to become inconsistent.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: updated, thx

dnduffy: Please push your updates.

src/Microsoft.ApplicationInsights.AspNetCore/project.json
yantang-msft: VS is so active in formatting things for you. :)

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: The AspNetCore's instrumentation code doesn't have null check for Request and Headers, and also our code for Request and Method, so I suppose it's guaranteed.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: Updated.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
yantang-msft: It's pushed. The update doesn't get reflected in this "outdated" conversation.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
pharring: Why public? Shouldn't it be private? And readonly?

yantang-msft: Can change to internal readonly. Internal here for test purpose.

PR #405, opened 2017-04-26
Fix OperationIdOfRequestIsPropagatedToChildDependency and TestBasicDependencyPropertiesAfterRequestingBasicPage functional tests

lmolkova:
1. OperationIdOfRequestIsPropagatedToChildDependency:
 - was failing because CorrelationIdLookupHelper was not set up properly and test was checking internal AI trace message instead on dependency telemetry
 - was failing because there are 2 dependencies: one is call to application from the test, another one was reported in the controller

2. TestBasicDependencyPropertiesAfterRequestingBasicPage 
was failing because AI was not disposed properly and magically application was using old telemetry channel remaining from the previous test and test was checking new telemetry channel.
While the magic that kept TelemetryChannel alive is still unknown, AI is now properly disposed and test is fixed.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/ApplicationInsightsInitializer.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/ApplicationInsightsInitializer.cs
@@ -43,7 +43,7 @@ namespace Microsoft.ApplicationInsights.AspNetCore
         /// </summary>
         public void Start()
         {
-            DiagnosticListener.AllListeners.Subscribe(this);
+            this.subscriptions.Add(DiagnosticListener.AllListeners.Subscribe(this));
         }
 
         /// <inheritdoc />
--- a/test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
+++ b/test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
@@ -1,4 +1,6 @@
-﻿namespace SampleWebAppIntegration.FunctionalTest
+﻿using System.Diagnostics;
+
+namespace SampleWebAppIntegration.FunctionalTest
 {
     using System;
     using System.Collections.Generic;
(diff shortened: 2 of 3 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/Implementation/ApplicationInsightsInitializer.cs
```diff
@@ -43,7 +43,7 @@ internal class ApplicationInsightsInitializer : IObserver<DiagnosticListener>, I
         /// </summary>
         public void Start()
         {
-            DiagnosticListener.AllListeners.Subscribe(this);
+            this.subscriptions.Add(DiagnosticListener.AllListeners.Subscribe(this));
```
yantang-msft: Seems like the magic should keep working since we are keep subscribing to the DiagnoticSource, but some how it doesn't work for that case?

test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
```diff
@@ -34,8 +37,15 @@ public void OperationIdOfRequestIsPropagatedToChildDependency()
...
-                Assert.Equal(telemetries[0].Context.Operation.Id, telemetries[1].Context.Operation.Id);
+                Assert.Equal(2, telemetries.Count(t => t.Context?.Operation?.Name != null && t.Context.Operation.Name.Contains(path)));
+                var dependency = (DependencyTelemetry)telemetries.Single(t =>
+                {
+                    var dt = t as DependencyTelemetry;
+                    return dt?.Context?.Operation?.Name != null && dt.Context.Operation.Name.Contains(path);
+                });
+
+                var request = (RequestTelemetry)telemetries.Single(t => t is RequestTelemetry);
+                Assert.Equal(dependency.Context.Operation.Id, request.Context.Operation.Id);
```
dnduffy: nitpick/question: I always wondered if the order for expected vs. actual should be reversed for semantic correctness even though it won't affect the outcome of the test?

lmolkova: Probably yes, since we expect dependency to have the same Id as request. I'll fix that

Issue #401, opened 2017-04-21
Hook up HttpDependenciesParsingTelemetryInitializer for dependencies parsing


SergeyKanzhelev:
Need to hook up the new telemetry initializer by default.

cijothomas:
This was done when dependecy module was added for .net core. closing.

PR #407, opened 2017-04-26
Enable dependency collector for .Net core framework

yantang-msft:
Enable dependency collector for .Net core framework and also fix unit tests

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
@@ -134,6 +134,7 @@
             services.AddSingleton<ITelemetryInitializer, WebSessionTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebUserTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, AspNetCoreEnvironmentTelemetryInitializer>();
+            services.AddSingleton<ITelemetryInitializer, HttpDependenciesParsingTelemetryInitializer>();
             services.AddSingleton<ITelemetryModule, DependencyTrackingTelemetryModule>();
 
 #if NET451
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
@@ -9,6 +9,7 @@
     using Microsoft.ApplicationInsights.AspNetCore.Extensions;
     using Microsoft.ApplicationInsights.AspNetCore.Logging;
     using Microsoft.ApplicationInsights.AspNetCore.TelemetryInitializers;
+    using Microsoft.ApplicationInsights.DependencyCollector;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.AspNetCore.Hosting;
     using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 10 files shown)


test/FunctionalTestUtils/TelemetryTestsBase.cs
```diff
@@ -80,13 +80,16 @@ public void ValidateBasicException(InProcessServer server, string requestPath, E
...
-#if NET451
         public void ValidateBasicDependency(string assemblyName, string requestPath, Func<IWebHostBuilder, IWebHostBuilder> configureHost = null)
         {
             DependencyTelemetry expected = new DependencyTelemetry();
             expected.ResultCode = "200";
             expected.Success = true;
+#if NET451
             expected.Name = requestPath;
+#else
+            expected.Name = "GET " + requestPath;
```
dnduffy: I thought these names were all normalized now, why is there a NET451 check?

test/FunctionalTestUtils/TelemetryTestsBase.cs
```diff
@@ -107,14 +110,20 @@ public void ValidateBasicDependency(string assemblyName, string requestPath, Fun
...
+
+            var dependencyTelemetry = dependencies.FirstOrDefault(d => d.Name == expected.Name
                   && d.Data == expected.Data
                   && d.Success == expected.Success
-                  && d.ResultCode == expected.ResultCode
-                );
+                  && d.ResultCode == expected.ResultCode);
+            Assert.NotNull(dependencyTelemetry);
+
+#if !NET451
```
dnduffy: Why !NET451?

yantang-msft: This is to test correlation, which doesn't work with NET451. It will work with NET46 though, and the code that make it work lives in the DiagnosticSource but only targeting NET46.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -134,10 +134,10 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
             services.AddSingleton<ITelemetryInitializer, WebSessionTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebUserTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, AspNetCoreEnvironmentTelemetryInitializer>();
+            services.AddSingleton<ITelemetryModule, DependencyTrackingTelemetryModule>();
```
lmolkova: please also add `HttpDependenciesParsingTelemetryInitializer ` (see #401 and https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/482 for motivation)

lmolkova:
I created #408: we need to test how AI work with AspNetCore2.0. This must also cover NET46 & Core 2.0.
There we can also make sure correlation works for NET46

test/FunctionalTestUtils/TelemetryTestsBase.cs
yantang-msft: No, it's not. I'll create an issue.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
yantang-msft: Added

test/FunctionalTestUtils/TelemetryTestsBase.cs
dnduffy: Per Sergey's comment on the issue, this is true for event source but in .NET Core aren't we using the diagnostic listener even for full framework?

lmolkova: if app is NET451 and uses AspNetCore, the HttpClient used is NET45 implementation, that could be instrumented with event source and profiler. Here we get only event source.

PR #411, opened 2017-04-28
Added encoding of the user identity being injected into the JavaScript.

dnduffy:
Added encoding of the user identity being injected into the JavaScript snippet and a corresponding unit test.


test/Microsoft.ApplicationInsights.AspNetCore.Tests/JavaScript/ApplicationInsightsJavaScriptTest.cs
```diff
@@ -11,19 +13,21 @@
 
     public static class ApplicationInsightsJavaScriptTest
     {
+        private static JavaScriptEncoder encoder = JavaScriptEncoder.Create(new UnicodeRange[] { UnicodeRanges.BasicLatin });
```
SergeyKanzhelev: is it what Chinese or Russian customers will expect? Which layer will decode it back to localized string?

dnduffy: That's not what this does.
It is saying anything which is not BasicLatin characters will get encoded, e.g. "<" becomes "\u003c" to prevent cross site scripting.
For the unit tests I am just setting a simple encoding.  For a live ASP.NET Core web application the framework will be automatically configured to some sensible defaults and users can always override the framework settings.

SergeyKanzhelev: So this is just a custom implementation for [HttpUtility.JavaScriptStringEncode](https://msdn.microsoft.com/en-us/library/dd991914.aspx)? Is it recommended way?

SergeyKanzhelev:
which package `JavaScriptStringEncoder` defined? Are we taking about MVC dependency for it?

test/Microsoft.ApplicationInsights.AspNetCore.Tests/JavaScript/ApplicationInsightsJavaScriptTest.cs
dnduffy: Yes that is how the documentation shows it being used.
Off the top of my head I don't know what the framework defaults that configuration to but I can look that up via debugging.  For the unit test this is all that is needed.

dnduffy:
There is no MVC dependency, [System.Text.Encodings.Web](https://www.nuget.org/packages/System.Text.Encodings.Web/) is a separate package on NuGet.

SergeyKanzhelev:
Should you modify `package.json` to add the reference then?

dnduffy:
@SergeyKanzhelev done.

Issue #192, opened 2016-05-10
Change success status to failed in case there's exception caught in RequestTrackingMiddleware.cs and response code is 200


kartang:

kartang:
Set the default status code in Exception Middleware, so that we make sure that status code is never <401 in case of exception, as we do in web.

Dmitry-Matveev:
If response to the client is 200 and the client is not affected, SDK would need to set request status to success and code to 200 just like ASP.NET core does. We should have exception telemetry item collected so that exception would not go unnoticed.

gzepeda:
Hi, 

We were looking for the file RequestTrackingMiddleware.cs to do the change but we are not able to find it. 

Is the context of the bug still valid or the file just got renamed? thanks!

dnduffy:
The AI SDK for .NET Core was rewritten a couple months ago to use listeners instead of middleware.
See HostingDiagnosticListener.cs.

gzepeda:
Thanks. We will check it out.

dnduffy:
I apologize, after looking at the pull request I believe this issue no longer existed after the rewrite and we just failed to close it.  The PR would actually create a bug where all status codes != 200 would suddenly report success which is not correct.

PR #415, opened 2017-04-29
Solved Issue #192

mazg-95:
Change success status to failed in case there's exception caught in RequestTrackingMiddleware.cs (HostingDiagnosticListener.cs) and response code is 200 #192 

https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/192


gzepeda:
We will be redoing this pull request without the commits that include the migration of the project to VS 2017 and only include the proposed changes. @mazg-95

Issue #416, opened 2017-04-30
v2.1.0-beta2 breaks WindowsAzure.Storage CreateIfNotExistsAsync


melvin2907:
Not sure if this is the correct repo to raise this issue however i reasonably certain that some code in the Application insights is the culprit. 

When upgrading from version 2.0.0 of package Microsoft.ApplicationInsights.AspNetCore to version v2.1.0-beta2 calls to the CreateIfNotExistsAsync method on the CloudQueue object in WindowsAzure.Storage. Start failing with the following error. 

Microsoft.WindowsAzure.Storage.StorageException: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.

henkmollema:
I'm seeing the same behavior after updating the NuGet package to 2.1.0-beta2.

michielpost:
Same issue here. I'm using `services.AddDataProtection().PersistKeysToAzureBlobStorage` from `Microsoft.AspNetCore.DataProtection.AzureStorage`, it throws the same StorageException when rendering a form with an AntiForgeryToken.

SergeyKanzhelev:
Hm, it seems that `DependencyTrackingTelemetryModule` was not initialized correctly. Azure queue should be exempt from headers modifications. See https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/198

I haven't tried this workaround, but I think you can get `DependencyTrackingTelemetryModule` from the list of services as `ITelemetryModule` and configure excluded domains. Something like this:

```
var modules = serviceProvider.GetServices<ITelemetryModule>();
var dependencyModule = services.FirstOrDefault<ServiceDescriptor>(t => t.ImplementationType == typeof(DependencyTrackingTelemetryModule));

var domains = dependencyModule .ExcludeComponentCorrelationHttpHeadersOnDomains;
domains.Add("core.windows.net");
domains.Add("core.chinacloudapi.cn");
domains.Add("core.cloudapi.de");
domains.Add("core.usgovcloudapi.net");
```

SergeyKanzhelev:
@yantang-msft @lmolkova can you please take a look and confirm that this is the case

maganuk:
I have the same issue. Went back to beta 1 and it works fine

DamianEdwards:
[...]

PR #418, opened 2017-05-01
Fix issue #416. Add certain domains to dependency collector's exclude list

yantang-msft:
Add some domains as a temporary workaround. Will probably be configurable when the configuration story is clear.

### DIFF
--- a/test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
+++ b/test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
@@ -20,38 +20,7 @@ namespace SampleWebAppIntegration.FunctionalTest
         private const string assemblyName = "MVCFramework45.FunctionalTests";
 
         [Fact]
-        public void OperationIdOfOutgoingRequestIsPropagated()
-        {
-#if !NET451 // Correlation is not supported in NET451. It works with NET46 and .Net core.
-            InProcessServer server;
-
-            using (server = new InProcessServer(assemblyName, InProcessServer.UseApplicationInsights))
-            {
-                using (var httpClient = new HttpClient())
-                {
-                    var task = httpClient.GetAsync(server.BaseHost + "/");
-                    task.Wait(TestTimeoutMs);
-                }
-            }
-
-            var telemetries = server.BackChannel.Buffer;
-            try
-            {
-                Assert.True(telemetries.Count >= 2);
-                var requestTelemetry = telemetries.OfType<RequestTelemetry>().Single();
-                var dependencyTelemetry = telemetries.OfType<DependencyTelemetry>().Single();
-                Assert.Equal(requestTelemetry.Context.Operation.Id, dependencyTelemetry.Context.Operation.Id);
-            }
-            catch (Exception e)
-            {
-                string data = DebugTelemetryItems(telemetries);
-                throw new Exception(data, e);
-            }
-#endif
-        }
-
-        [Fact]
-        public void OperationIdOfOutgoingRequestIsNotPropagatedIfDomainIsExcluded()
+        public void CorrelationInfoIsNotAddedToRequestHeaderIfUserAddDomainToExcludedList()
         {
 #if !NET451 // Correlation is not supported in NET451. It works with NET46 and .Net core.
             InProcessServer server;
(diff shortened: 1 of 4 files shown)


test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/ApplicationInsightsExtensionsTests.cs
```diff
@@ -314,7 +314,7 @@ public static void RegistersTelemetryConfigurationFactoryMethodThatPopulatesItWi
                 Assert.Equal(1, modules.Count());
 #endif
 
-                var dependencyModule = services.FirstOrDefault<ServiceDescriptor>(t => t.ImplementationType == typeof(DependencyTrackingTelemetryModule));
+                var dependencyModule = services.FirstOrDefault<ServiceDescriptor>(t => t.ImplementationFactory?.GetMethodInfo().ReturnType == typeof(DependencyTrackingTelemetryModule));
```
SergeyKanzhelev: can you please add a test to validate that the default list of excluded domains is not empty. I wonder whether we can have a unit test that works with the queue and validates that we didn't break the authentication by our headers. Without exposing Azure queue access key

yantang-msft: I added the test to make sure the correlation header is not added if the domain is in the exclude list. To validate the authentication is not broken, since the validation happens at server side, we may have to at least provide a fake key to create a storage client. Although the case I added is not storage specific, it somehow helps.

test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
```diff
@@ -11,12 +11,80 @@ namespace SampleWebAppIntegration.FunctionalTest
...
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.DependencyCollector;
+    using Microsoft.Extensions.DependencyInjection;
 
     public class DependencyTelemetryMvcTests : TelemetryTestsBase
     {
         private const string assemblyName = "MVCFramework45.FunctionalTests";
 
         [Fact]
+        public void OperationIdOfOutgoingRequestIsPropagated()
```
SergeyKanzhelev: this supposed to be tested by Dependency Collector Module itself. The only test needed in this repo is that everything was configured correctly

test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
```diff
@@ -11,12 +11,80 @@ namespace SampleWebAppIntegration.FunctionalTest
...
+            catch (Exception e)
+            {
+                string data = DebugTelemetryItems(telemetries);
+                throw new Exception(data, e);
+            }
+#endif
+        }
+
+        [Fact]
+        public void OperationIdOfOutgoingRequestIsNotPropagatedIfDomainIsExcluded()
```
SergeyKanzhelev: rename test - it tests that configuration of the module can be updated by customer and this update will be used

test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
yantang-msft: It makes sense to me at first. But to test the E2E cross component correlation functionality, the DependencyCollector and the listener (this project) are both needed, so it's hard to say it's whose job. Since it's just a test case, I'll leave it here.

test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
yantang-msft: Renamed.

test/MVCFramework45.FunctionalTests/FunctionalTest/DependencyTelemetryMvcTests.cs
yantang-msft: Moved to CorrelationMVCTests.cs

PR #419, opened 2017-05-02
Make UseAI and AddAI calls idempotent

pakrym:
To avoid duplicated telemetries if `AddApplicationInsightsTelemetry` was called while `UseAddApplicationInsights` is in `Program.Main`

/cc @SergeyKanzhelev @Dmitry-Matveev @dnduffy


DamianEdwards:
This is required for the Azure light-up scenarios to work correctly and thus is needed in 2.1.0

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -269,5 +274,11 @@ internal static void AddTelemetryConfiguration(IConfiguration config, Applicatio
                 serviceOptions.ApplicationVersion = version;
             }
         }
+
+        private static bool IsApplicationInsightsAdded(IServiceCollection services)
+        {
+            // We treat ApplicationInsightsInitializer as a marker that AI services were addede to service collection
```
henkmollema: small typo: `addede`

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerCallbacks.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace Microsoft.ApplicationInsights.AspNetCore.Logging
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Class to control default debug logger and disable it if logger was added to <see cref="ILoggerFactory"/> explicetely.
```
pharring: typo: explicitly

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerCallbacks.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    internal class ApplicationInsightsLoggerCallbacks
+    {
+        /// <summary>
+        /// Stores list of callbacks to execute when new ApplicationInsights logger is added.
+        /// </summary>
+        private readonly List<Action> callbacks;
```
pharring: You just re-invented MulticastDelegate, Just use an event:
```C#
event EventHandler LoggerAdded;
```
and let clients subscribe.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerCallbacks.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Registers a callback that executes when new ApplicationInsights logger is added.
+        /// </summary>
+        public void AddLoggerCallback(Action loggerAddedCallback)
+        {
+            lock (callbacks)
+            {
+                foreach (var callback in callbacks)
+                {
```
pharring: Should this really be in the same place where you register new loggers?
Shouldn't there be two methods: e.g. AddLoggerCallback and NotifyLoggers?
It looks like we've just re-invented a MulticastDelegate. Perhaps an event would be better?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
```diff
@@ -50,11 +50,27 @@ public static ILoggerFactory AddApplicationInsights(this ILoggerFactory factory,
...
+            return factory.AddApplicationInsights(serviceProvider, filter, () => { });
+        }
+
+        /// <summary>
+        /// Adds an ApplicationInsights logger that is enabled as defined by the filter function.
+        /// </summary>
+        /// <param name="factory"></param>
+        /// <param name="filter"></param>
+        /// <param name="serviceProvider">The instance of <see cref="IServiceProvider"/> to use for service resolution.</param>
+        /// <param name="loggerAddedCallback">The callback that get's executed when another ApplicationInsights logger is added.</param>
```
pharring: typo: "gets" should not have an apostrophe.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -269,5 +274,11 @@ internal static void AddTelemetryConfiguration(IConfiguration config, Applicatio
...
+            return services.Any(service => service.ServiceType == typeof(ApplicationInsightsInitializer));
```
pharring:
```C#
return services.OfType<ApplicationInsightsInitializer>().Any();
```

pakrym: It's not what it does.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerCallbacks.cs
pharring: Also, the description no longer makes sense for this class.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
```diff
@@ -50,11 +50,27 @@ public static ILoggerFactory AddApplicationInsights(this ILoggerFactory factory,
             IServiceProvider serviceProvider,
             Func<string, LogLevel, bool> filter)
         {
+            return factory.AddApplicationInsights(serviceProvider, filter, () => { });
```
pharring: Can't you use 'null' instead of creating an empty delegate? Then it can be a default parameter. Handle null in ```AddLoggerCallback``` with ```callback?.Invoke()```

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerCallbacks.cs
```diff
@@ -0,0 +1,38 @@
...
+    internal class ApplicationInsightsLoggerCallbacks
+    {
+        /// <summary>
+        /// Stores list of callbacks to execute when new ApplicationInsights logger is added.
+        /// </summary>
+        private readonly List<Action> callbacks;
+
+        public ApplicationInsightsLoggerCallbacks()
+        {
+            callbacks = new List<Action>();
```
pharring: Could be inline with the field definition, then the constructor evaporates.

Dmitry-Matveev:
It looks like this fix will make UseAI and AddAI idempotent in the case on single-threaded execution.
Is there a way AddAI might be invoked from the several threads at the same time and still end up with two things configured because both threads will pass the check on "IsEnabled"?

pakrym:
@Dmitry-Matveev we are not concerned with multi threaded scenario here, configuration part of application is inherently one threaded.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -122,45 +124,48 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
...
-            services.AddSingleton<ITelemetryInitializer, ApplicationInsights.AspNetCore.TelemetryInitializers.OperationCorrelationTelemetryInitializer>();
-            services.AddSingleton<ITelemetryInitializer, SyntheticTelemetryInitializer>();
-            services.AddSingleton<ITelemetryInitializer, WebSessionTelemetryInitializer>();
-            services.AddSingleton<ITelemetryInitializer, WebUserTelemetryInitializer>();
-            services.AddSingleton<ITelemetryInitializer, AspNetCoreEnvironmentTelemetryInitializer>();
-            services.AddSingleton<ITelemetryInitializer, HttpDependenciesParsingTelemetryInitializer>();
-            services.AddSingleton<ITelemetryModule, DependencyTrackingTelemetryModule>();
+            if (!IsApplicationInsightsAdded(services))
+            {
+                services.TryAddSingleton<IHttpContextAccessor, HttpContextAccessor>();
```
SergeyKanzhelev: Does the error from this method should be traced?

PR #420, opened 2017-05-02
Solved Issue #192 #415 

mazg-95:
Change success status to failed in case there's exception caught in RequestTrackingMiddleware.cs (HostingDiagnosticListener.cs) and response code is 200 #192

Contributors:
- gzepeda
- oscargarciacolon
- Cesar64garcia
- mazg-95
- gabrielaHerreraPoggio
- luisfermata1
- estuardo9609
- JRosales0912
- joshy110

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
@@ -285,12 +285,16 @@
                 var telemetry = httpContext?.Features.Get<RequestTelemetry>();
                 if (telemetry != null)
                 {
-                    telemetry.Success = false;
+                    if (int.TryParse(telemetry.ResponseCode, out int _responseCode))
+                    {
+                        telemetry.Success = !(_responseCode == 200);
+                    }
                 }
 
                 var exceptionTelemetry = new ExceptionTelemetry(exception);
                 exceptionTelemetry.HandledAt = ExceptionHandledAt.Platform;
                 exceptionTelemetry.Context.GetInternalContext().SdkVersion = this.sdkVersion;
+
                 this.client.Track(exceptionTelemetry);
             }
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -285,12 +285,16 @@ private void OnException(HttpContext httpContext, Exception exception)
                 var telemetry = httpContext?.Features.Get<RequestTelemetry>();
                 if (telemetry != null)
                 {
-                    telemetry.Success = false;
+                    if (int.TryParse(telemetry.ResponseCode, out int _responseCode))
+                    {
+                        telemetry.Success = !(_responseCode == 200);
```
dnduffy: So if the response code is 200 then you set telemetry.Success to false but otherwise you set it to true for all other response codes?  Isn't that almost the opposite of the existing logic?

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -285,12 +285,16 @@ private void OnException(HttpContext httpContext, Exception exception)
                 var telemetry = httpContext?.Features.Get<RequestTelemetry>();
                 if (telemetry != null)
                 {
-                    telemetry.Success = false;
+                    if (int.TryParse(telemetry.ResponseCode, out int _responseCode))
```
dnduffy: nitpick: per coding standards for C# please don't use underscores as prefixes to variable names.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
dnduffy: After reviewing the original issue and this PR more closely I believe that this logic is a bug and that the original issue in #192 no longer existed.

dnduffy:
Thank-you for your contribution, I'm sorry we didn't realize the rewrite contained the fix and close issue #192 sooner.

PR #422, opened 2017-05-03
Removed clearing entire nuget cache from dirs.proj as this is clearin…

cijothomas:
…g entire machine cache in build servers and slowing down total time.


cijothomas:
@SergeyKanzhelev 3.5 minutes for aspnet core build. (more than that, this is clearing nuget cache from entire buildmachine which is also used by other builds who also experience this delay)

Issue #344, opened 2017-02-06
[Docs / Guidance] RE: Custom ITelemetryProcessor


ctolkien:
Just flagging that your own custom `ITelemetryProcessors` will be instantiated, but the `Process` method will  _never_ be called if your `IntrumentationKey` is left blank. 

For instance, in the scenario where you are developing locally and trying to get your own `ITelemetryProcessor` to work.

Yes, I just spent my morning trying to figure this one out...  🥇

Dracs:
Oh, thank you for that. I was pulling my hair out over this. Adding in the `IntrumentationKey` value got my code being called.

ohadschn:
Why would this be a documentation / guidance issue? 
In other words, why not call `Process` even when `IntrumentationKey` is blank?

SergeyKanzhelev:
I think it may be addressed with the new multi-sink support work https://github.com/Microsoft/ApplicationInsights-dotnet/issues/559

SergeyKanzhelev:
Addressed with #604.

PR #445, opened 2017-05-15
Ability to add telemetry processors through DI

karolz-ms:
This is necessary to complete exception snapshotting work for .NET Core and overall a good addition, filling a gap in programmatic initialization capabilities of AI SDK.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
@@ -1,6 +1,7 @@
 namespace Microsoft.Extensions.DependencyInjection
 {
     using System;
+    using System.Linq;
     using System.Collections.Generic;
     using Microsoft.ApplicationInsights.AspNetCore.Extensions;
     using Microsoft.ApplicationInsights.Channel;
@@ -21,6 +22,7 @@ namespace Microsoft.Extensions.DependencyInjection
         private readonly IEnumerable<ITelemetryInitializer> initializers;
         private readonly IEnumerable<ITelemetryModule> modules;
         private readonly ITelemetryChannel telemetryChannel;
+        private readonly IEnumerable<Func<ITelemetryProcessor, ITelemetryProcessor>> telemetryProcessorFactories;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="T:TelemetryConfigurationOptionsSetup"/> class.
@@ -29,11 +31,13 @@ namespace Microsoft.Extensions.DependencyInjection
             IServiceProvider serviceProvider,
             IOptions<ApplicationInsightsServiceOptions> applicationInsightsServiceOptions,
             IEnumerable<ITelemetryInitializer> initializers,
-            IEnumerable<ITelemetryModule> modules)
+            IEnumerable<ITelemetryModule> modules,
+            IEnumerable<Func<ITelemetryProcessor, ITelemetryProcessor>> telemetryProcessorFactories)
         {
             this.applicationInsightsServiceOptions = applicationInsightsServiceOptions.Value;
             this.initializers = initializers;
             this.modules = modules;
+            this.telemetryProcessorFactories = telemetryProcessorFactories;
             this.telemetryChannel = serviceProvider.GetService<ITelemetryChannel>();
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -45,6 +49,15 @@ public void Configure(TelemetryConfiguration configuration)
                 configuration.InstrumentationKey = this.applicationInsightsServiceOptions.InstrumentationKey;
             }
 
+            if (this.telemetryProcessorFactories.Count() > 0)
```
pharring: Use ```this.telemetryProcessorFactories.Any()``` (it short-circuits the enumeration once it hits the first item)

karolz-ms: Good point

pharring: Actually, perhaps this if statement is unnecessary. If the imported ```telemetryProcessorFactories``` is empty, then the foreach does nothing, but I think we should still call ```.Build()``` for the case where users have manually added processors to the ```TelemetryProcessorChainBuilder```. Someone needs to call ```.Build()``` and this might be the right place.

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -45,6 +49,15 @@ public void Configure(TelemetryConfiguration configuration)
                 configuration.InstrumentationKey = this.applicationInsightsServiceOptions.InstrumentationKey;
             }
 
+            if (this.telemetryProcessorFactories.Any())
```
pharring: Actually, I wonder if we should remove the if statement altogether and let it fall through to the (potentially empty) ```foreach``` followed by ```.Build()```. In the case where users have manually added to the ```TelemetryProcessorChainBuilder```, someone needs to call ```.Build()``` and this might be the most appropriate place.

karolz-ms: I think whoever adds manually to the `TelemetryProcessorChainBuilder` should also take care of calling `Build()`. We do, so we call. Calling `Build()` "just in case" is a little dubious. The original code did not do that and I did not want to change this.

pharring: OK.

SergeyKanzhelev: BTW, quick pulse telemetry module and telemetry processor need to know about each other. This link between them will not survive thru the re-`Build()`. So `Initialize` on module should be called after `Build` on the chain.

Can this be achieved in this case?

(real solution will be to change the design of quick pulse)

dnduffy: @karolz-ms and @pharring Would we have a "race" condition where someone could add a processor factory after we call Build()?  Something like the issue with calling both UseApplicationInsights and AddApplicationInsightsTelemetry?

karolz-ms: @SergeyKanzhelev The code that this change is adding runs before quick pulse telemetry module and processor are added/initialized. So I believe we are fine here?

I agree that quick pulse implementation should probably be changed. This dependency between its telemetry module and telemetry processor is making thins unnecessarily fragile. Do you want me to open an issue for this?

@dnduffy In general TelemetryProcessorChainBuilder.Use() is just adding a TP factory to the chain builder. The Build() call is what instantiates the chain. Each time Build() is called the chain is instantiated from scratch, or at least this is the current implementation. So it is fine to Use(TP factory), then call Build(), then Use(another TP factory), then call Build() again. That is--IF there are no dependencies like quick pulse has today. But that is a separate issue that I am not trying to address here.

karolz-ms:
@dnduffy 
1. Not sure what default setup might be missing... For AddTlemetryChannelAndProcessorsForFullFramework, the TP factories that that method adds will be added to the same TelemetryProcessorChainBuilder, so the user factories and our predefined factories will be used simultaneously to build the final chain.

2. Nope, as I explained above, each call to Build() will result in a brand-new chain. The old chain will be discarded, see https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/Extensibility/Implementation/TelemetryProcessorChainBuilder.cs#L71

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
dnduffy: I just realized what I was thinking of, it wasn't the Use/Build of the processor chain but the fact that internal helper methods were causing the re-adding of the extra processors etc.

dnduffy:
Relevant to #344.

Issue #447, opened 2017-05-16
A TelemetryProcessor that is also a TelemetryModule is not Initialized


pharring:
#445 introduced a way to set up TelemetryProcessors via DI.
If a TelemetryProcessor _also_ implements ```ITelemetryModule``` the base SDK will call ```Initialize``` on it.
We should do the same in aspnetcore.

pharring:
@karolz-ms FYI

pharring:
As a general comment, trying to replicate all the behaviors of the base SDK in this repo is a losing proposition. Can we consider exposing some of the internals of the telemetry pipeline (particularly the configuration factory) from the Base SDK - either by making them public or adding an "InternalsVisibleTo" from the base SDK to this one?

karolz-ms:
@pharring So do you think we should implement this behavior to unblock SnapshotCollector while using this issue to track the more comprehensive resolution?

pharring:
@karolz-ms yes, I think so.

PR #448, opened 2017-05-17
Telemetry processors are initialized if they implement ITelemetryModule

karolz-ms:
This change is required to unblock exception snapshotting on .NET Core, for more info see issue https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/447

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
@@ -81,6 +81,15 @@ namespace Microsoft.Extensions.DependencyInjection
             {
                 module.Initialize(configuration);
             }
+
+            foreach (ITelemetryProcessor processor in configuration.TelemetryProcessors)
+            {
+                ITelemetryModule module = processor as ITelemetryModule;
+                if (module != null)
+                {
+                    module.Initialize(configuration);
+                }
+            }
         }
 
         private void AddTelemetryChannelAndProcessorsForFullFramework(TelemetryConfiguration configuration)
--- a/test/Microsoft.ApplicationInsights.AspNetCore.Tests/FakeTelemetryProcessor.cs
+++ b/test/Microsoft.ApplicationInsights.AspNetCore.Tests/FakeTelemetryProcessor.cs
@@ -16,14 +16,14 @@
             }
 
             this.next = next;
-            this.Initialized = false;
+            this.IsInitialized = false;
         }
 
-        public bool Initialized { get; private set; }
+        public bool IsInitialized { get; private set; }
 
         public void Initialize(TelemetryConfiguration configuration)
         {
-            this.Initialized = true;
+            this.IsInitialized = true;
         }
 
         public void Process(ITelemetry item)
--- a/test/Microsoft.ApplicationInsights.AspNetCore.Tests/FakeTelemetryProcessor.cs
+++ b/test/Microsoft.ApplicationInsights.AspNetCore.Tests/FakeTelemetryProcessor.cs
@@ -4,7 +4,7 @@
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
 
-    internal class FakeTelemetryProcessor : ITelemetryProcessor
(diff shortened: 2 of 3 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -81,6 +81,15 @@ public void Configure(TelemetryConfiguration configuration)
             {
                 module.Initialize(configuration);
             }
+
+            foreach (ITelemetryProcessor processor in configuration.TelemetryProcessors)
+            {
+                ITelemetryModule module = processor as ITelemetryModule;
```
dnduffy: So I udnerstand the situation where a class implements both ITelemetryProcessor and ITelemetryModule but why wouldn't the ITelemetryModule also be registered so that it is handled by the existing foreach above?

pharring:
@dnduffy In the base SDK, if you're using a .config file, then any TelemetryProcessor that implements ITelemetryModule is automatically Initialized without needing to be mentioned in the TelemetryModules section. As I mentioned in #447 I think it would be wise not to deviate from established behavior when you're using DI. Indeed, I'd like it if DI just called into helper methods in the base SDK. So, to answer your question directly: It is needed to maintain consistency. A class that implements both could be injected twice, but try explaining that to users (which ones need to be injected twice?).

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
karolz-ms: @dnduffy Not really, as Paul alluded below, this change is intended for telemetry processors that also require initialization (that is why they implement ITelemetryModule). They won't be _registered_ as telemetry modules.

karolz-ms:
@dnduffy There is also a simpler reason: as a telemetry processor, the class must be instantiated by the framework (TelemetryProcessorChainBuilder), because it needs the "next processor" instance. So it cannot be registered as telemetry module.

BTW IMO this is a flaw in AI SDK object model. It would be much more desirable if telemetry processors did not know about the next one in the chain and were not responsible for calling it. Instead, it should be the framework who is responsible for calling the processors and passing telemetry records to them. This would allow much more flexibility, both at initialization time, as well as at run time (parallel execution etc). But I guess changing this would be a breaking change and we are not ready to go down that path.

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
dnduffy: It feels wrong to use ITelemetryModule to get an Initialize method if it isn't really a telemetry module.

karolz-ms: Ehm, from https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/Extensibility/ITelemetryModule.cs

```csharp
/// <summary>
    /// Represents an object that supports initialization from <see cref="TelemetryConfiguration"/>.
    /// </summary>
```
That is pretty much all there is to it :)

dnduffy: Maybe I'm carrying semantic baggage from past usage in my head. ;)

dnduffy:
@pharring I agree it is confusing which is why I find the re-use of the module interface confusing.
@karolz-ms I would probably prefer it wasn't a true chain also, but it is.
@SergeyKanzhelev any thoughts on changing processors to not be explicitly chained?

test/Microsoft.ApplicationInsights.AspNetCore.Tests/FakeTelemetryProcessor.cs
```diff
@@ -16,6 +16,14 @@ public FakeTelemetryProcessor(ITelemetryProcessor next)
             }
 
             this.next = next;
+            this.Initialized = false;
```
dnduffy: nitpick: "IsInitialized".

karolz-ms:
@dnduffy Thanks for the review! Can we get this change merged then?

PR #483, opened 2017-06-27
Use ExceptionTelemetry for tracking exceptions

### LINKED ISSUE #393: Logging exceptions with ExceptionTelemetry
When you log an error with an exception through the `ILogger` interface like this:
```cs
logger.Log(LogLevel.Error, 0, state, exception, messageFormatter);
```
[it will use a `TraceTelemetry` instance](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/5aa38ae65469aa649a50ad5433ae6d65d5b4e406/src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLogger.cs#L44).

Is it possible to use an `ExceptionTelemetry` instance in the case where an exception is present? Exception telemetries provide a nicer experience in UI's such as Visual Studio and the Azure portal.

henkmollema:
Tracking `ExceptionTelemetry` instances rather than `TraceTelemetry` provides a nicer experience in UI's such as Visual Studio and the Azure portal.

Resolves #393

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLogger.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using Microsoft.ApplicationInsights;
+    using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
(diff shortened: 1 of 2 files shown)


henkmollema:
Thanks @dnduffy! Any chance this makes it into 2.1?

dnduffy:
@henkmollema We actually had the final build for 2.1 on Wednesday and just released it today after a final test pass.  However, there is talk of pushing a couple features into a 2.1.1 before 2.2 so it might not wait very long for inclusion in a stable release.

henkmollema:
I see, no worries. Thanks!

PR #486, opened 2017-06-30
Add alias to AI logger provider

pakrym:
In logging 2.0 we added ability to configure logging filters in centralized way. To configure filters per logging provider provider full name should be used or alternatively alias specified by `ProviderAliasAttribute`.

`ProviderAliasAttribute` is late bound and is detected by full type name so no reference to new aspnetcore packages is required.

@dnduffy @SergeyKanzhelev

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/ProviderAliasAttribute.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/ProviderAliasAttribute.cs
@@ -1,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/Implementation/ProviderAliasAttribute.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
dnduffy: Should we have the standard MS copyright header instead?

pakrym:
I removed copyright from this file, an one other file it was accidentally added to in the past.

pakrym:
Can we ship build to the feed when this change goes in, so we can start E2E testing?

dnduffy:
@pakrym Yes, we can make the builds auto-upload to myget, usually we only do the upload when we want to make something available (such as this change you requested) but we should automate it like some of the other builds do.

Issue #482, opened 2017-06-27
No default instance for type Microsoft.ApplicationInsights.Extensibility.ITelemetryProcessor when using 3rd party IoC/DI container.


f1xxxer:
When App Insights package is added to project which is using 3rd party IoC/DI container (StructureMap in my case) it is falling with error `StructureMapConfigurationException: No default instance or named instance 'Default' for requested plugin type Microsoft.ApplicationInsights.Extensibility.ITelemetryProcessor`

Solution is to add your own custom dummy processor and then everything works, but i assume that shouldn't be the default behavior for the package?

My environment:

Microsoft Visual Studio Professional 2017 Preview (2)
Version 15.3.0 Preview 2.1

<PropertyGroup>
    <TargetFramework>netcoreapp1.1</TargetFramework>
    <RuntimeFrameworkVersion>1.1.1</RuntimeFrameworkVersion>
    <PackageTargetFallback>$(PackageTargetFallback);dotnet5.6;dnxcore50;portable-net45+win8</PackageTargetFallback>
      .........
      .........
  </PropertyGroup>
<PackageReference Include="Microsoft.AspNetCore.Mvc" Version="1.1.2" />

paule96:
Hi Guys,

we have the same problem with our project. This is only a Problem in the newest Visual Studio preview.
If we use the console with dotnet run it will works fine. 

Maybe it's help to find the problem.

DamianEdwards:
Interesting. Does this only happen when launching in VS under the debugger? Do you have a reference to the AppInsights SDK package in your project? Could you share your project's CSPROJ file here?

kirkone:
Hi,

here is an example .csproj:

```
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <Folder Include="wwwroot\" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Diagnostics" Version="2.0.0-preview2-final" />
    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.0.0-preview2-final" />
    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.0.0-preview2-final" />
[...]

Issue #430, opened 2017-05-09
Support setting request operation name based on executing Razor Page


DamianEdwards:
Currently, the request telemetry doesn't construct a friendly operation name for requests served by Razor Pages in ASP.NET Core 2.0. The [`MvcDiagnosticsListener`](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/develop/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/MvcDiagnosticsListener.cs) should be updated to grab details of the current executing page (if indeed it's a page) and use that to construct the request operation name, e.g. /Pages/Index.cshtml

pakrym:
Why do you want page file name instead of it's action descriptor display name? (`/Index` in your case)

DamianEdwards:
Good question. What do you think folks would find more useful? That's what it really comes down to.

pakrym:
In case of pages having file name is nice, but it's very inconsistent with what we do for views.

DamianEdwards:
But I think that makes sense given the different way the route info is constructed, i.e. from the page file.

dnduffy:
Personally I would expect to see the route that a user sees rather than the backing implementation details but I don't know if that view is representative of other users.

pakrym:
With file name operation name would look like `GET /Pages/Index.cshtml` and it looks like user requested actual file vs `GET /Index` if we use route values.

I implemented the route values way to start with, mostly because it's simpler and more consistent.

DamianEdwards:
Fair enough.

PR #489, opened 2017-07-06
Fix Func injection issue on structuremap and track MVC page names

pakrym:
Instead of injecting `Func<,>` factory interface is introduced.

OperationName from MVC Razor Page is generated the same way controller/action is makes for simpler code change and consistent behavior.

https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/482
https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/430

@dnduffy

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
@@ -3,6 +3,7 @@ namespace Microsoft.Extensions.DependencyInjection
     using System;
     using System.Linq;
     using System.Collections.Generic;
+    using Microsoft.ApplicationInsights.AspNetCore;
     using Microsoft.ApplicationInsights.AspNetCore.Extensions;
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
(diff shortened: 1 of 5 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -51,9 +52,9 @@ public void Configure(TelemetryConfiguration configuration)
 
             if (this.telemetryProcessorFactories.Any())
             {
-                foreach (Func<ITelemetryProcessor, ITelemetryProcessor> processorFactory in this.telemetryProcessorFactories)
+                foreach (ITelemetryProcessorFactory processorFactory in this.telemetryProcessorFactories)
```
dnduffy: What about existing TelemetryProcessors that were created the old way?  This seems like something that should go into 2.2 and not 2.1.1.

dnduffy: I think @karolz-ms and @pharring are using it.
I agree it is a good change, I just don't think it belongs in this "hotfix" release of bug fixes.

karolz-ms: This will be a small, but breaking change for the exception exemplification feature.

pakrym: If this fix is not shipped with AspNetCore 2.0, any application using structure map would be broken because we inject AI everywhere by default.

karolz-ms: What is "structure map"?

pakrym: @karolz-ms AspNetCore allows usage of third party DI containers if they support common basic set of features. http://structuremap.github.io/ is one of these containers.

pharring: Snapshot Collector ("exception exemplification") doesn't reference this repo. It references the base repo. This won't be a breaking change for us. We'll just have to update our docs.

Issue #488, opened 2017-07-01
Fix for issue #416 does not apply to Azure Storage Emulator


joergjo:
Please see #416. SDK 2.1 still breaks Azure Storage access when using the emulator.

PR #490, opened 2017-07-10
Fix issue 488.

cijothomas:
And Version updates

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - [Address the issue where DependencyCollection breaks Azure Storage Emulator calls](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/488)
 - [Support setting request operation name based on executing Razor Page](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/430)
 - [Fixed ITelemetryProcessor dependency injection failure when using 3rd party IoC Container](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/482)
+- [Logging exceptions when using ILogger if an exception is present](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/393)
 - Updated SDK version dependency to 2.4.1 for DependencyCollector.
 
 ## Version 2.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Version 2.1.1
+
+- [Address the issue where DependencyCollection breaks Azure Storage Emulator calls](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/488)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Version 2.1.1
+
+- [Address the issue where DependencyCollection breaks Azure Storage Emulator calls](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/488)
+- [Support setting request operation name based on executing Razor Page](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/430)
+- [Fixed ITelemetryProcessor dependency injection failure when using 3rd party IoC Container](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/482)
+- Updated SDK version dependency to 2.4.1 for DependencyCollector.
+
```
dnduffy: Please check the history and add in at least the following items (but with better descriptions than I provided here):
- Updated the ILogger logging of exceptions to send ExceptionTelemetry instead of TraceTelemetry.
- Added sanitization of authenticated user name to the JavaScript snippet output.
- Anything else that changed in develop since 2.1.0.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -146,6 +146,8 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
                     excludedDomains.Add("core.chinacloudapi.cn");
                     excludedDomains.Add("core.cloudapi.de");
                     excludedDomains.Add("core.usgovcloudapi.net");
+                    excludedDomains.Add("localhost");
+                    excludedDomains.Add("127.0.0.1");
```
dnduffy: Are we sure we want these always excluded?  It was mentioned in the issue that this will break tracking local dependencies when a developer sets up everything on a single machine for local development and testing.  I don't think this is something that should be universally forced on everyone.  I know a lot of developers used to run a "onebox" with all of their services and databases installed locally when testing a component or service that was part of a larger SOA system.  This configuration would exclude tracking such a development setup.

cijothomas: Yes, this is the same change as made in Web as well. We don't want to break customers by default. Developers using other services in local machine, and remove this to monitor those calls. (not perfect either, but better than the experience of all emulator calls failing by default)

cijothomas:
Wont merge until functional tests are failed due to this change

cijothomas:
Fix: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/488
Related: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/416

PR #492, opened 2017-07-10
Synchronize HTTP context access

pakrym:
To prevent HTTP context corruption

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/TelemetryInitializerBase.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/TelemetryInitializerBase.cs
@@ -33,15 +33,18 @@
                 return;
             }
 
-            var request = context.Features.Get<RequestTelemetry>();
-
-            if (request == null)
+            lock (context)
             {
-                AspNetCoreEventSource.Instance.LogTelemetryInitializerBaseInitializeRequestNull();
-                return;
-            }
+                var request = context.Features.Get<RequestTelemetry>();
 
-            this.OnInitializeTelemetry(context, request, telemetry);
+                if (request == null)
+                {
+                    AspNetCoreEventSource.Instance.LogTelemetryInitializerBaseInitializeRequestNull();
+                    return;
+                }
+
+                this.OnInitializeTelemetry(context, request, telemetry);
+            }
         }
 
         protected abstract void OnInitializeTelemetry(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - [Support setting request operation name based on executing Razor Page](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/430)
 - [Fixed ITelemetryProcessor dependency injection failure when using 3rd party IoC Container](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/482)
 - [Logging exceptions when using ILogger if an exception is present](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/393)
+- [Syncronize access to HttpContext properties](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/373)
 - Updated SDK version dependency to 2.4.1 for DependencyCollector.
 
 ## Version 2.1.0


src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/TelemetryInitializerBase.cs
```diff
@@ -33,15 +33,18 @@ public void Initialize(ITelemetry telemetry)
                 return;
             }
 
-            var request = context.Features.Get<RequestTelemetry>();
-
-            if (request == null)
+            lock (context)
```
SergeyKanzhelev: this is a partial solution and it will not work when customer will try to access `RequestTelemetry` himself. As we discussed in the issue - the best solution will be to move `RequestTelemetry` from features back to DI and make sure it has most of the properties initialized as early as possible.

Even the first step will make the possibility of a crash way less possible.

SergeyKanzhelev: It is a change of a behavior since we are locking a public class. Please mention in `CHANGELOG.md` as part of PR.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/TelemetryInitializerBase.cs
```diff
@@ -33,15 +33,18 @@ public void Initialize(ITelemetry telemetry)
...
             {
-                AspNetCoreEventSource.Instance.LogTelemetryInitializerBaseInitializeRequestNull();
-                return;
-            }
+                var request = context.Features.Get<RequestTelemetry>();
```
SergeyKanzhelev: is [this](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/ead65975f0fa27650b40983265bad05fdea16a95/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/MvcDiagnosticsListener.cs#L24) call safe? Should it be locked as well? Any chance customer code running in parrellel with the MVC OnBegin?

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/TelemetryInitializerBase.cs
```diff
@@ -33,15 +33,18 @@ public void Initialize(ITelemetry telemetry)
...
 
-            this.OnInitializeTelemetry(context, request, telemetry);
+                if (request == null)
+                {
+                    AspNetCoreEventSource.Instance.LogTelemetryInitializerBaseInitializeRequestNull();
+                    return;
+                }
+
+                this.OnInitializeTelemetry(context, request, telemetry);
```
SergeyKanzhelev: potentially we can reduce the lock down to the immediate call to `HttpContext`. Most of telemetry initializers are not doing anything when the property is already set. So we wasting the `lock` on every cached item

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/TelemetryInitializerBase.cs
pakrym: Failure to retrieve `RequestTelemetry` from `Features` is just a first symptom of parallel `HttpContext` access so fixing it would not really fix anything, context has multiple collections inside that are not thread safe.

SergeyKanzhelev: Yes, but moving it to DI will make us avoid locking. Concurrent dictionary works faster than explicit lock on retrieving the element from regular dictionary.

davidfowl:
> please do not lock for cached items. If you think it's feasible to move RequestTelemetyr from features to DI - I think it should be done (unless ASP.NET team has plans to make features collection thread safe).

That doesn't solve the problem though... The goal with this PR was to make a minimal change that has the biggest bang for the buck. Keep in mind that this *is* a per request lock and won't content most of the time as there's no parallelism in the 90% case.

> Beside I still have concern that this is not the best implementation. However it is the best sop gap solution we can come up with now. So I'll approve for the version we about to ship, but I'd like a better solution going forward. Please do not close the issue.

Agreed. We think that copying the properties is a better solution (see the begin/end API I suggested) but this is a stop gap. We should meet in person and discuss the overall solution.

SergeyKanzhelev:
@davidfowl makes sense. I'd like however even for this change to move the lock down the exact location http context being accessed so cached properties will not require to take an extra lock. @pakrym can you do this change soon?

@cijothomas worst case if @pakrym cannot do it now we can merge as is and address perf optimization later.

cijothomas:
i'll wait till about 4pm and merge.. I plan to send signed bits tonight to manual validation, and publish to nuget.org by morning.

cijothomas:
@pakrym - please add this to changelog.md as well. Thanks.

PR #499, opened 2017-07-15
Add net46 target to source, and replace testing with net46 instead of net451

cijothomas:
Added a new target net46 for the aspnet core project, and replaced all testing to target net46 instead of net451. The motivation behind this is to run tests using stable version of xunit, which no longer support net451.


cijothomas:
Need to try signed builds to make sure signing is not broken/missing due to this change.

PR #500, opened 2017-07-15
Add an AddApplicationInsightsTelemetryProcessor extension method

pharring:
We changed the way TelemetryProcessors are injected in #489 
This makes it more cumbersome for callers, since they have to define a `TelemetryProcessorFactory` class instead of just using an inline lambda.

We can help by adding a generic extension method that uses a generic telemetry processor factory.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
@@ -181,6 +181,19 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         /// <summary>
+        /// Adds an Application Insights Telemetry Processor into a service collection via a <see cref="ITelemetryProcessorFactory"/>.
+        /// </summary>
+        /// <typeparam name="T">Type of the telemetry processor to add.</typeparam>
+        /// <param name="services">The <see cref="IServiceCollection"/> instance.</param>
+        /// <returns>
+        /// The <see cref="IServiceCollection"/>.
+        /// </returns>
+        public static IServiceCollection AddApplicationInsightsTelemetryProcessor<T>(this IServiceCollection services) where T : ITelemetryProcessor
+        {
+            return services.AddSingleton<ITelemetryProcessorFactory>(new TelemetryProcessorFactory<T>());
+        }
+
+        /// <summary>
         /// Adds Application Insight specific configuration properties to <see cref="IConfigurationBuilder"/>.
         /// </summary>
         /// <param name="configurationSourceRoot">The <see cref="IConfigurationBuilder"/> instance.</param>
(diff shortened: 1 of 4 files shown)


pharring:
Waiting for a review. @dnduffy @pakrym @karolz-ms since you've all dabbled in this space.

This change would simplify the sample code I had to add at https://docs.microsoft.com/en-us/azure/application-insights/app-insights-snapshot-debugger#configure-snapshot-collection-for-aspnet-core-20-applications

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryProcessorFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+    internal class TelemetryProcessorFactory<T> : ITelemetryProcessorFactory where T : ITelemetryProcessor
+    {
+        /// <summary>
+        /// Creates an instance of the telemetry processor, passing the
+        /// next <see cref="ITelemetryProcessor"/> in the call chain to
+        /// its constructor.
+        /// </summary>
+        public ITelemetryProcessor Create(ITelemetryProcessor next)
+        {
+            return (ITelemetryProcessor)Activator.CreateInstance(typeof(T), args: next);
```
pakrym: Use  ActivatorUtilities (https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Extensions.DependencyInjection.Abstractions/DIActivatorUtilities.cs#L12) so `ITelemetryProcessor` can inject other services in constructor.

pharring: Thanks for the tip. Will update.

pakrym:
Tests?

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -181,6 +181,19 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
...
 
         /// <summary>
+        /// Adds an Application Insights Telemetry Processor into a service collection via a <see cref="ITelemetryProcessorFactory"/>.
+        /// </summary>
+        /// <typeparam name="T">Type of the telemetry processor to add.</typeparam>
+        /// <param name="services">The <see cref="IServiceCollection"/> instance.</param>
+        /// <returns>
+        /// The <see cref="IServiceCollection"/>.
+        /// </returns>
+        public static IServiceCollection AddApplicationInsightsTelemetryProcessor<T>(this IServiceCollection services) where T : ITelemetryProcessor
```
pakrym: What about extension methods that take instance or `Type`?

pharring: Are you suggesting we add a non-generic overload that takes a Type parameter? What's the benefit?

pakrym: Usually we tend to have non generic overload for cases you don't have generic `T` (for example you are searching for types using reflection).

pharring: OK, I added a non-generic overload.

SergeyKanzhelev:
@cijothomas it looks good. Please merge

pharring:
Actually, don't merge just yet. I want to add a unit test to cover the DI case.

pharring:
(That's the 2nd time I've accidentally hit "Close and comment" instead of "Comment". Grr.)

pharring:
I added the extra unit test. This is ready to merge now.

PR #517, opened 2017-08-16
upgrade to the 2.5.0-beta1 and enable Adaptive sampling on all ptatfo…

SergeyKanzhelev:
…rms. Tests are failing

Using latest version of Application Insights (2.5.0-beta1) exposed the problem with multi-sink. Unit tests failing as the channel cannot be overridden on the DefaultSink.

@karolz-ms do you have ides off the top of your head? Otherwise I'll keep debugging in the end of the day

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Version 2.2.0-beta1
 
 - Project is upgraded to work with Visual Studio 2017. Also projects are modified to use csproj instead of project.json.
+- Adaptive sampling enabled for both - full framework and .NET Core applications.
 
 ## Version 2.1.1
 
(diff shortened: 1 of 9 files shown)


karolz-ms:
@SergeyKanzhelev no, sorry. I looked briefly at the errors but the call stack(s) do not have an obvious connection with `TelemetrySink` feature. Looks like you are already further down the road of investigating the cause.

I also do not quite get the part about "channel cannot be overridden on default sink". It should be--e.g. test `TelemetryChannelCanBeSetByUserToReplaceDefaultChannelForTesting` https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationTest.cs#L292 is supposed to verify that.

SergeyKanzhelev:
@karolz-ms it seems to be some behavior related to the order of `Build` call for the telemetry processors chain. As I said I'll keep debugging. I saw that the channel we assign to the configuration wasn't used in default sink - `InMemoryChannel` was used instead.

SergeyKanzhelev:
@karolz-ms working theory - `TelemetryProcessorSink` got initialized when default sink is created, but never updated when the new channel being used later

Issue #524, opened 2017-09-08
Error in getting cross-component correlationid?


andrasg:
I have a weird exception on one of my servers. It occurs at the second request after appstart. It crashes the appdomain completely.

```
System.InvalidOperationException: Invalid non-ASCII or control character in header: 0x000A
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.FrameHeaders.ThrowInvalidHeaderCharacter(Char ch)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.FrameHeaders.ValidateHeaderCharacters(String headerCharacters)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.FrameHeaders.ValidateHeaderCharacters(StringValues headerValues)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.FrameResponseHeaders.SetValueFast(String key, StringValues value)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.FrameHeaders.Microsoft.AspNetCore.Http.IHeaderDictionary.set_Item(String key, StringValues value)
   at Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners.HttpHeadersUtilities.SetHeaderKeyValue(IHeaderDictionary headers, String headerName, String keyName, String keyValue)
   at Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners.HostingDiagnosticListener.SetAppIdInResponseHeader(HttpContext httpContext, RequestTelemetry requestTelemetry)
   at Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners.HostingDiagnosticListener.OnBeginRequest(HttpContext httpContext, Int64 timestamp)
   at Proxy_Method_From_<>f__AnonymousType0`2_To_Void OnBeginRequest(Microsoft.AspNetCore.Http.HttpContext, Int64)(Object , Object , IProxyFactory )
   at Microsoft.Extensions.DiagnosticAdapter.DiagnosticSourceAdapter.Write(String diagnosticName, Object parameters)
   at Microsoft.Extensions.DiagnosticAdapter.DiagnosticSourceAdapter.System.IObserver<System.Collections.Generic.KeyValuePair<System.String,System.Object>>.OnNext(KeyValuePair`2 value)
   at System.Diagnostics.DiagnosticListener.Write(String name, Object value)
[...]

PR #525, opened 2017-09-08
Check appId response code before using content

andrasg:
This pullrequest is to solve an issue where appId detection is incorrectly using a non-200 http response as appId and is breaking applications.

Issue: #524


cijothomas:
Thanks @andrasg for reporting the issue and PR for fix.
Will merge after running builds.

andrasg:
@cijothomas, I was wondering if I can easily add tests to verify this scenario, but it would require major rework (providing HttpClient/WebRequest from outside to be able to pipe in a mock).
Do you want me to add tests?

PR #526, opened 2017-09-08
Added functional tests for 2.0

cijothomas:


test/EmptyApp20.FunctionalTests/FunctionalTest/RequestTelemetryEmptyAppTests.cs
```diff
@@ -0,0 +1,72 @@
+﻿namespace EmptyApp20.FunctionalTests.FunctionalTest
+{
+    using System.Linq;
+    using System.Net.Http;
+    using FunctionalTestUtils;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Xunit;
```
Dmitry-Matveev: It looks like Sergey was removing XUnit with [this PR](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/608/files), so we might want to use MSTEST here instead as well.

test/FunctionalTestUtils20/FunctionalTestUtils20.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <VersionPrefix>1.0.2</VersionPrefix>
+    <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
+    <DelaySign>true</DelaySign>
+    <PreserveCompilationContext>true</PreserveCompilationContext>
+    <AssemblyName>FunctionalTestUtils20</AssemblyName>
+    <AssemblyOriginatorKeyFile>../../keys/35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>true</SignAssembly>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <PackageId>FunctionalTestUtils</PackageId>
+    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
```
Dmitry-Matveev: Should it be 1.6.1 or something like 2.0?

cijothomas: Good catch! 2.0.

test/EmptyApp20.FunctionalTests/FunctionalTest/RequestTelemetryEmptyAppTests.cs
cijothomas: The issue faced in other repo is not applicable in aspnet core repro as we don't target anything below net451. 
We can replace it in future if need arises, but as of now i dont see strong reason to remove it. It'll require some work as the output logging is based on xunit now.

PR #530, opened 2017-09-21
Enable ServerTelemetryChannel in .net core by the default

iusafaro:
Enable ServerTelemetryChannel in .net core by the default


test/MVCFramework20.FunctionalTests/FunctionalTest/TelemetryModuleWorkingMvcTests.cs
```diff
@@ -17,7 +18,18 @@ public TelemetryModuleWorkingMvcTests(ITestOutputHelper output) : base(output)
         [Fact]
         public void TestBasicDependencyPropertiesAfterRequestingBasicPage()
         {
-            this.ValidateBasicDependency(assemblyName, "/Home/About/5", InProcessServer.UseApplicationInsights);
+            const string RequestPath = "/Home/About/5";
+
+            using (var server = new InProcessServer(assemblyName))
```
cijothomas: Is there a reason of moving this line (using (var server = new InProcessServer(assemblyName))) from base class to each of the subclasses?

test/FunctionalTestUtils20/InProcessServer.cs
```diff
@@ -18,22 +16,28 @@ public class InProcessServer : IDisposable
...
 
         public InProcessServer(string assemblyName, Func<IWebHostBuilder, IWebHostBuilder> configureHost = null)
         {
             this.configureHost = configureHost;
+
             var machineName = Environment.GetEnvironmentVariable("COMPUTERNAME");
             this.url = "http://" + machineName + ":" + random.Next(5000, 14000).ToString();
-            this.backChannel = this.Start(assemblyName);
+
+            this.listener = new TelemetryHttpListenerObservable("http://localhost:4001/v2/track/");
```
cijothomas: Lets make this a constant.

cijothomas:
Due to some reason build is not triggered automatically! Can you trigger one against this PR. ?

test/MVCFramework20.FunctionalTests/FunctionalTest/TelemetryModuleWorkingMvcTests.cs
iusafaro: It was in base class only for dependencies, but for requests and exceptions, it was out of it in individual cases - i moved it out of here for dependencies as well to be more consistent in all methods.

KEMBL:
Hello guys, 

thank you for enabling of ServerTelemetryChannel in NETCOREAPP2_0, but how to access it now?

in non-Core it is possible like this:

		var channel = new Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.ServerTelemetryChannel();
		channel.StorageFolder = "./CacheFolder/";
		channel.Initialize(TelemetryConfiguration.Active);


how to get the same in core?

cijothomas:
@KEMBL 
Can you open an GH issue, so that the question gets monitored.

Can you retrieve channel like this hostingEngine.Services.GetService<ITelemetryChannel>()) ?

Issue #569, opened 2017-12-12
Support EventId in ApplicationInsightsLogger


duracellko:
It would be nice, if `ApplicationInsightsLogger` would include `EventId.Id` and `EventId.Name` in telemetry properties ad EventId and EventName.

older:
What is official way to enable EventId today? I used documentation from [https://github.com/Microsoft/ApplicationInsights-aspnetcore/wiki/Logging](here), but it is now deprecated and recommended replacement `Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerOptions` does not even have `IncludeEventId` property.

cijothomas:
@older In the new provider Microsoft.Extensions.Logging.ApplicationInsightsLoggingProvider, EventID and EventName are always included. You dont need to do anything extra to enable it.

Please give it a try, and if EventId/Name are not captured, let us know.
https://docs.microsoft.com/en-us/azure/azure-monitor/app/ilogger

PR #570, opened 2017-12-12
Include EventId in telemetry properties

duracellko:
Fix issue #569.

Main update is to `ApplicationInsightsLogger`. It includes EventId and EventName properties in telemetry. It is included only, when EventId has non-default values.
By default this functionality is off for compatibility. Following example can switch it on:
This code should be ASP.NET Core application Startup class.
```C#
public IServiceProvider ConfigureServices(IServiceCollection services)
{
    services.Configure<ApplicationInsightsLoggerOptions>(Configuration.GetSection("ApplicationInsights"));
    services.AddLogging(logging =>
        {
            logging.AddConfiguration(Configuration.GetSection("Logging"))
                .AddConsole()
                .AddDebug();
            logging.Services.TryAddEnumerable(ServiceDescriptor.Singleton<ILoggerProvider, ApplicationInsightsLoggerProvider>());
        });
}
```

Additionally test coverage of ApplicationInsightsLogger was improved.


cijothomas:
@duracellko Thanks for your contribution!
We are currently fixing some issues related to symbols, and the next beta release.
Will look at this and merge as soon as the next beta is released.
Thanks!

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - Live metrics collection is enabled by default for .NET Core applications (was already enabled for full .NET applications).
 - Updated Web/Base SDK version dependency to 2.5.0-beta1.
 - DependencyCollector referred from 2.5.0-beta1 supports collecting SQL dependency calls in .NET Core Applications using EntityFramework.
+- ApplicationInsightsLogger adds EventId into telemetry properties. It is off by default for compatibility. It can be switched on by configuring ApplicationInsightsLoggerOptions.
```
cijothomas: Can you move this to 2.3.0-beta1 please?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerProvider.cs
```diff
@@ -21,10 +25,20 @@ public ApplicationInsightsLoggerProvider(TelemetryClient telemetryClient, Func<s
             this.filter = filter;
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ApplicationInsightsLoggerProvider"/> class.
+        /// </summary>
+        public ApplicationInsightsLoggerProvider(TelemetryClient telemetryClient, IOptions<ApplicationInsightsLoggerOptions> options)
+        {
+            this.telemetryClient = telemetryClient;
+            this.filter = trueFilter;
```
cijothomas: Why use a true filter here?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerOptions.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Microsoft.ApplicationInsights.AspNetCore.Logging
+{
+    /// <summary>
+    /// Options for configuration of <see cref="ApplicationInsightsLogger"/>.
+    /// </summary>
+    public class ApplicationInsightsLoggerOptions
```
cijothomas: There is another PR which introduces ```ApplicationInsightsLoggerOptions```
https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/574/files
Can you also have took at that as well?

I think passing ```ApplicationInsightsLoggerOptions``` to both ApplicationInsightsLoggerProvider and ApplicationInsightsLogger as done in the other PR is a better option. Can you incorporate this change to your PR?

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Logging/ApplicationInsightsLoggerTests.cs
```diff
@@ -30,6 +30,174 @@ public void TestLoggerSetsSdkVersionOnLoggedTelemetry()
             Assert.True(isCorrectVersion);
         }
 
+        /// <summary>
```
cijothomas: Thanks a lot for adding sufficient unit tests!

duracellko:
Thank you for comments. I will have a look at those later this week.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerOptions.cs
duracellko: Thank you for the hint. Yes, it's actually better. Also I realized that classes are internal, so I can change signature of constructors.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerProvider.cs
duracellko: I followed pattern in [ConsoleWriter](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs), but it is not needed with new approach.

PR #620, opened 2018-03-08
add fxcop analyzer

MS-TimothyMothra:
Add the FxCop Analyzer to Non-Test projects.
This changed number an analyzer warnings from 304 to 320.
I fixed the 2 nameof() warnings.
New total 218


MS-TimothyMothra:
@cijothomas do you know why the results would show up in Warnings instead of Errors?

PR #622, opened 2018-03-09
Prevent reuse and dispose of TelemetryConfiguration by multiple web hosts

lmolkova:
Fix Issue https://github.com/Microsoft/ApplicationInsights-dotnet/issues/613 .
<Short description of the fix.>
Creates a new instance of telemetry configuration by each Web host and updates TelemteryConfiguration.Active on the first run
- [x] I ran Unit Tests locally.

- [x] Changes in public surface reviewed
- [ ] Design discussion issue #
- [x] CHANGELOG.md updated with one line description of the fix, and a link to the original issue.
- [ ] The PR will trigger build, unit tests, and functional tests automatically. If your PR was submitted from fork - mention one of committers to initiate the build for you.

@cijothomas could you please trigger tests for me and add me to the contributors list  in this repo?

- Please follow [these] (https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/develop/Readme.md) instructions to build and test locally.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
@@ -6,7 +6,7 @@
     using Microsoft.Extensions.Options;
 
     /// <summary>
-    /// The <see cref="IOptions{TelemetryConfiguration}"/> implementation that create new <see cref="TelemetryConfiguration"/> every time when called"/>.
+    /// The <see cref="IOptions{TelemetryConfiguration}"/> implementation that create new <see cref="TelemetryConfiguration"/> every time when called".
     /// </summary>
     internal class TelemetryConfigurationOptions : IOptions<TelemetryConfiguration>
     {
@@ -21,9 +21,8 @@
                 c.Configure(this.Value);
             }
 
-            // as we expect some customers to use TelemetryConfiguration.Active even together with DependencyInjection
-            // we make sure it has been set up
-            // it must be done only once even if there are multiple Web Hosts in the process
+            // as we expect some customers to use TelemetryConfiguration.Active together with dependency injection
+            // we make sure it has been set up, it must be done only once even if there are multiple Web Hosts in the process
             if (!IsActiveConfigured(this.Value.InstrumentationKey))
             { 
                 foreach (var c in configureOptionsArray)
@@ -37,11 +36,11 @@
         public TelemetryConfiguration Value { get; }
 
         /// <summary>
-        /// Determines is TelemetryConfiguration.Active needs to be configured
+        /// Determines if TelemetryConfiguration.Active needs to be configured.
         /// </summary>
-        /// <param name="instrumentationKey">Instrumentation key configuration</param>
-        /// <returns>True is Active TelemertryConfiguration was previously configured</returns>
-        private bool IsActiveConfigured(string instrumentationKey)
+        /// <param name="instrumentationKey">Instrumentation key.</param>
+        /// <returns>True is TelemertryConfiguration.Active was previously configured.</returns>
+        private static bool IsActiveConfigured(string instrumentationKey)
         {
             var active = TelemetryConfiguration.Active;
             if (string.IsNullOrEmpty(active.InstrumentationKey) && !string.IsNullOrEmpty(instrumentationKey))
@@ -49,7 +48,7 @@
                 return false;
             }
 
-            if (active.TelemetryInitializers.Count <= 1 && active.TelemetryProcessors.Count <= 1)
+            if (active.TelemetryInitializers.Count <= 1)
             {
                 return false;
             }
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
@@ -10,9 +10,10 @@
     /// </summary>
     internal class TelemetryConfigurationOptions : IOptions<TelemetryConfiguration>
     {
+        private readonly object lockObject = new object();
+
         public TelemetryConfigurationOptions(IEnumerable<IConfigureOptions<TelemetryConfiguration>> configureOptions)
         {
-            // workaround for Microsoft/ApplicationInsights-dotnet#613
             this.Value = TelemetryConfiguration.CreateDefault();
 
             var configureOptionsArray = configureOptions.ToArray();
@@ -21,13 +22,17 @@
                 c.Configure(this.Value);
             }
 
-            // as we expect some customers to use TelemetryConfiguration.Active together with dependency injection
-            // we make sure it has been set up, it must be done only once even if there are multiple Web Hosts in the process
-            if (!IsActiveConfigured(this.Value.InstrumentationKey))
-            { 
-                foreach (var c in configureOptionsArray)
+            lock (lockObject)
+            {
+                // workaround for Microsoft/ApplicationInsights-dotnet#613
+                // as we expect some customers to use TelemetryConfiguration.Active together with dependency injection
+                // we make sure it has been set up, it must be done only once even if there are multiple Web Hosts in the process
+                if (!IsActiveConfigured(this.Value.InstrumentationKey))
                 {
-                    c.Configure(TelemetryConfiguration.Active);
+                    foreach (var c in configureOptionsArray)
+                    {
+                        c.Configure(TelemetryConfiguration.Active);
+                    }
                 }
             }
         }
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
@@ -1,28 +1,60 @@
 ﻿namespace Microsoft.Extensions.DependencyInjection
 {
     using System.Collections.Generic;
+    using System.Linq;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.Extensions.Options;
 
     /// <summary>
-    /// The <see cref="IOptions{TelemetryConfiguration}"/> implementation that uses <see cref="TelemetryConfiguration.Active"/> as initial value.
+    /// The <see cref="IOptions{TelemetryConfiguration}"/> implementation that create new <see cref="TelemetryConfiguration"/> every time when called"/>.
     /// </summary>
     internal class TelemetryConfigurationOptions : IOptions<TelemetryConfiguration>
     {
         public TelemetryConfigurationOptions(IEnumerable<IConfigureOptions<TelemetryConfiguration>> configureOptions)
         {
             // workaround for Microsoft/ApplicationInsights-dotnet#613
-            this.Value = new TelemetryConfiguration();
-            this.Value.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
+            this.Value = TelemetryConfiguration.CreateDefault();
 
-            foreach (var c in configureOptions)
+            var configureOptionsArray = configureOptions.ToArray();
+            foreach (var c in configureOptionsArray)
             {
                 c.Configure(this.Value);
-                c.Configure(TelemetryConfiguration.Active);
+            }
+
+            // as we expect some customers to use TelemetryConfiguration.Active even together with DependencyInjection
+            // we make sure it has been set up
+            // it must be done only once even if there are multiple Web Hosts in the process
+            if (!IsActiveConfigured(this.Value.InstrumentationKey))
+            { 
+                foreach (var c in configureOptionsArray)
+                {
+                    c.Configure(TelemetryConfiguration.Active);
+                }
             }
         }
 
         /// <inheritdoc />
         public TelemetryConfiguration Value { get; }
+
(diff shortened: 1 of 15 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
```diff
@@ -1,23 +1,60 @@
...
+                }
+            }
         }
 
         /// <inheritdoc />
-        public TelemetryConfiguration Value => TelemetryConfiguration.Active;
+        public TelemetryConfiguration Value { get; }
+
+        /// <summary>
+        /// Determines is TelemetryConfiguration.Active needs to be configured
```
pharring: if (not is)

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
```diff
@@ -1,23 +1,60 @@
...
+            // workaround for Microsoft/ApplicationInsights-dotnet#613
+            this.Value = TelemetryConfiguration.CreateDefault();
+
+            var configureOptionsArray = configureOptions.ToArray();
+            foreach (var c in configureOptionsArray)
             {
                 c.Configure(this.Value);
             }
+
+            // as we expect some customers to use TelemetryConfiguration.Active even together with DependencyInjection
```
pharring: Nit: "Dependency Injection" (two words)

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
```diff
@@ -1,23 +1,60 @@
...
 
         /// <inheritdoc />
-        public TelemetryConfiguration Value => TelemetryConfiguration.Active;
+        public TelemetryConfiguration Value { get; }
+
+        /// <summary>
+        /// Determines is TelemetryConfiguration.Active needs to be configured
+        /// </summary>
+        /// <param name="instrumentationKey">Instrumentation key configuration</param>
+        /// <returns>True is Active TelemertryConfiguration was previously configured</returns>
```
pharring: if (not is)

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
```diff
@@ -1,23 +1,60 @@
...
+        private bool IsActiveConfigured(string instrumentationKey)
+        {
+            var active = TelemetryConfiguration.Active;
+            if (string.IsNullOrEmpty(active.InstrumentationKey) && !string.IsNullOrEmpty(instrumentationKey))
+            {
+                return false;
+            }
+
+            if (active.TelemetryInitializers.Count <= 1 && active.TelemetryProcessors.Count <= 1)
```
pharring: I just want to check: Does TelemetryProcessors.Count include the always-present final processor that sends the telemetry item to the channel? If so, cool. Otherwise, surely it's possible to have an active telemetry configuration with no initializers and no processors. It might be unusual for most ASP.NET Core apps, but still possible.

lmolkova: Yes, this default processor is set up when the TelemetryConfiguration.Active is constructed.
Perhaps checking for processors count does not make sense. Likely it is still one even after configuration.

This code attempts to configure Active static singleton in the same way as it configured TelemetryConfiguration created for DI. The latter could not possibly have 1 TelemetryIntializer (at least at this point, initializers could be removed later, but now they have been just configured) - this method is called after all telemetry initializers are added and there are like 20 of them.

Note that also we don't really support the case of having multiple different configurations in the same process. And this PR does not really help with it. It just does some damage control for the co-existing TelemetryConfiguration.Active and TelemetryConfiguration DI singleton.

If a user manually modifies Active and removes everything AND runs multiple webhosts, then he would need to update Active in every WebHost he creates. That sounds like a weird scenario, though.

lmolkova:
Build failure 
```
2018-03-09T20:47:37.4170034Z Chunk '1' attempt '2' of file 'drop/Microsoft.ApplicationInsights.AspNetCore.2.2.1.nupkg' fail to send request to server. Error: Microsoft.VisualStudio.Services.Common.VssUnauthorizedException: TF400813: The user '***' is not authorized to access this resource.
```

perhaps I don't have permissions to publish the build at this point, i.e. not a code/test issue

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
```diff
@@ -1,23 +1,59 @@
...
+        /// <returns>True is TelemertryConfiguration.Active was previously configured.</returns>
+        private static bool IsActiveConfigured(string instrumentationKey)
+        {
+            var active = TelemetryConfiguration.Active;
+            if (string.IsNullOrEmpty(active.InstrumentationKey) && !string.IsNullOrEmpty(instrumentationKey))
+            {
+                return false;
+            }
+
+            if (active.TelemetryInitializers.Count <= 1)
```
Dmitry-Matveev: This looks like a new enforced condition that we should have at least 2 telemetry initializers for Active to be configured. Should we add a test to actually require it (default TI > 2) and don't miss if some TIs are removed later?

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
```diff
@@ -1,23 +1,59 @@
...
+            this.Value = TelemetryConfiguration.CreateDefault();
+
+            var configureOptionsArray = configureOptions.ToArray();
+            foreach (var c in configureOptionsArray)
             {
                 c.Configure(this.Value);
             }
+
+            // as we expect some customers to use TelemetryConfiguration.Active together with dependency injection
+            // we make sure it has been set up, it must be done only once even if there are multiple Web Hosts in the process
```
Dmitry-Matveev: If couple `TelemetryConfigurationOptions` objects can be constructed in parallel, then we may still try to configure `Active` twice due to positive concurrency in the code. Do we expect such behavior or do we need to protect against it all? (I feel like I'm imagining things here..)

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/ApplicationInsightsExtensionsTests.cs
```diff
@@ -119,20 +118,21 @@ public static void RegistersTelemetryConfigurationFactoryMethodThatReadsInstrume
             }
 
             /// <summary>
-            /// Tests that the Active configuration singleton is used as the telemetry configuration instance by the configuration factory.
-            /// This demonstrates that existing documentation for how to create a telemetry client and track custom events etc. works in ASP.NET 5
-            /// when no ApplicationInsights.config file exists but a project.json file does exist which contains the instrumentation key.
+            /// Tests that the Active configuration singleton is updated, but another instance of telemetry configuration is created for dependency injection.
+            /// ASP.NET Core developers should always use Dependency Injection instead of static singleton approach.
```
Dmitry-Matveev: It's been a pattern for a while in AI SDKs to use `Active` for central overrides such as instrumentation key or channels. Should we update some documentation in response to this change to make sure it does not contradict the new behavior?

test/WebApi20.FunctionalTests/FunctionalTest/MultipleWebHostsTests.cs
```diff
@@ -0,0 +1,140 @@
...
+                Assert.Single(telemetry2.Where(IsServiceDependencyCall));
+                Assert.DoesNotContain(telemetry2, t => t is TelemetryItem<ExceptionData>);
+
+                var request2 = telemetry2.Single(t => t is TelemetryItem<RequestData>);
+                Assert.Equal("200", ((TelemetryItem<RequestData>)request2).data.baseData.responseCode);
+            }
+        }
+
+        [Fact]
+        public void TwoWebHostsOneIsDisposed()
```
Dmitry-Matveev: Do we need a case for both being disposed? `Active` may still stay with its default channel on, will it hurt anything?

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
lmolkova: the other TI is coming from base SDK, so enforcing it with tests in aspnetcore SDK would not really help.

However if it ever changes, tests in aspnetcore SDK would fail for some other issues and we'd probably have to spend a lot of time figuring out why. I'd add a test with a comment to address this

src/Microsoft.ApplicationInsights.AspNetCore/Implementation/TelemetryConfigurationOptions.cs
lmolkova: it's very unlikely, but still possible, so it's a valid concern. I'll add some sync code around. This happens normally once per web host during service resolution, so should not be a performance issue

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/ApplicationInsightsExtensionsTests.cs
lmolkova: The difference in the behavior is only noticeable when altering TelemetryConfiguration.Active after calling `AddApplicationInsights`.

Previously, adding TelemetryInitializers/Processors, modifying other settings would have an effect on the telemetry collected automatically by the modules: requests, depednencies, exceptions.

Now, it would not as TelemetryConfiguration used by SDK and Active are different objects.

I think it's just a best practice for DI-based systems to never use static singleton and always resolve services through the DI.

We also have documentation that shows how to properly do things in AspNetCore SDK here: https://docs.microsoft.com/en-us/azure/application-insights/app-insights-api-filtering-sampling

However still, it could be a breaking change as some of apps may follow other not-that-good practices.

test/WebApi20.FunctionalTests/FunctionalTest/MultipleWebHostsTests.cs
lmolkova: Active will stay, as Active lifetime is a process lifetime. It's a good idea to test it still works after all web hosts are gone as in ServiceFabric SharedProcess mode, some background tasks are not bound to any web host and run before and after web hosts are created.

cijothomas:
@lmolkova I have given you contributor access to this repo.
 I will review this code tomorrow. ( I was OOF for last week)

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/ApplicationInsightsExtensionsTests.cs
cijothomas: @lmolkova 
We have been advising people (incorrectly?) to get TelemetryConfiguration from Active, and modify it in some of the documentations.
Example below:
https://github.com/Microsoft/ApplicationInsights-aspnetcore/wiki/Custom-Configuration#adding-new-telemetryinitializer
"one can obtain the TelemetryConfiguration object (either from DI or by using TelemetryConfiguration.Active), and add new initializers."
The correct way to modify TelemetryConfiguration so that SDK modules also pick up changes is to get TelemetryConfiguration from DI, and modify it - is this correct? I can update the wiki docs to reflect this as the correct approach.

cijothomas: If this change introduce potential issues for folks who relied on SDK modules picking up TC.Active modifications, lets mention that in Changelog.md itself so its easier to explain later :)

cijothomas:
Some issues in build system is preventing me from queueing Linux build. Will check this and reply.
Have triggered a regular windows build anyway.
(Due to some VSTS security issue, builds are not triggered now for forked repos. This is expected to be fixed in April)

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/ApplicationInsightsExtensionsTests.cs
lmolkova: @cijothomas 
It is correct that DI approach should be always preferred over static singletons in ASP.Net Core or any project where DI is used.  
I.e. 
`TelemetryConfiguration.Active.TelemetryInitializers.Add(new MyCustomTelemetryInitializer())`

should not be promoted as a way to add telemetry initializers.
The reasons are demonstrated by this bug: issues global state, hidden dependencies, poor testability.

And you are right, it introduces potential issues for those who used TC.Active. I'll update changelog with this info.

cijothomas: Can you please look at this FAQ i added and make corrections if needed ?
https://github.com/Microsoft/ApplicationInsights-aspnetcore/wiki/FAQs#2-accessing-telemetryconfiguration-via-telemetryconfigurationactive-vs-using-di

lmolkova: @cijothomas looks great, thanks for changing docs!

Issue #817, opened 2018-12-21
Deprecate older ILogger implementation in package and move to newer implementation


RamjotSingh:
A newer implementation of the ILogger is now pulled into a separate package in PR https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/239

The current implementation in ASPNetCore package needs to be deprecated and guidance needs to be updated to ask people to move to the newer implementation.

- Mark current ILogger implementation in this repo as deprecated
- Decide if ASPNetCore package takes a dependency on the new ILogger package or we directly ask developers to install both. 
- Based on decision above we might need to mark the existing extension method in ASPNetCore package as deprecated.

PR #818, opened 2019-01-03
Deprecate ILogger extensions in favor of Microsoft.Extensions.Logging.ApplicationInsights package

lmolkova:
Fix Issue #817.

ApplicationInsights Logger is now implemented in Microsoft.Extensions.Logging.ApplicationInsights nuget package.

Please install the new package and configure logging using `ApplicationInsightsLoggingBuilderExtensions.AddApplicationInsights` overloads.

See examples in https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/239

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Version 2.6.0-beta3
+- [Deprecate ApplicationInsightsLoggerFactoryExtensions.AddApplicationInsights logging extensions in favor of Microsoft.Extensions.Logging.ApplicationInsights package](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/817)
 - [Fix: Do not track requests by each host in the process](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/621)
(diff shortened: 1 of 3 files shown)


lmolkova:
cc @RamjotSingh

RamjotSingh:
Shouldn't the classes implementing ILogger be marked as Obselete too? Files under 
https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/develop/src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation

cijothomas:
Those are internal. So no need to mark obsolete really.

RamjotSingh:
Also we need a good strategy for the longer run (3.x). Are we going to ask users to install both packages or are we going to introduce a dependency?

lmolkova:
> Also we need a good strategy for the longer run (3.x). Are we going to ask users to install both packages or are we going to introduce a dependency? 

@cijothomas  and @Dmitry-Matveev  decided to ask users to install both packages. 
I think based on the feedback we may introduce a dependency.

We should document new logger (https://github.com/Microsoft/ApplicationInsights-dotnet-logging/issues/251) and consider changing VS templates.

PR #852, opened 2019-03-20
Remove DiagnosticAdapter and use DiagnosticListener directly

lmolkova:
Try out using DiagnosticListener without DiagnosticAdapter and see it it improves performance

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
@@ -1,4 +1,6 @@
-﻿namespace Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners
+﻿using System.Collections.Generic;
+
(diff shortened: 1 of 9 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs
```diff
@@ -559,5 +562,79 @@ public void Dispose()
         {
             SubscriptionManager.Detach(this);
         }
+
+        public void OnNext(KeyValuePair<string, object> value)
+        {
+            if (value.Key == "Microsoft.AspNetCore.Hosting.HttpRequestIn.Start")
```
pharring: This would be better as a switch instead of an if/else/../else chain.

lmolkova:
I think @Dmitry-Matveev has proof that it indeed saves some performance and it should be merged.
Current functional tests are sufficient.

I believe the main reason for perf hit from DiagnosticsAdapter was in IRouteData interface - we created this to avoid taking the dependency on the  Microsoft.AspNetCore.Routing.Abstractions and DiagnosticAdapter was dynamically typing IRouteData from that package to ours.

Issue #603, opened 2018-02-16
VS Debugger/Azure Web Apps - application insights logger LogLevel is not set correctly


cijothomas:
If you are reporting bug/issue, please provide detailed Repro instructions.

## Repro Steps
1.  Create new Asp.Net Core Web App from VS default template. ( I picked WebApi)
2.  Configure AI ikey. (useai or addai extension method)
3. Start App from Visual Studio (debugging), fire requests

or
Just create a new app, make NO changes and publish to Azure WebApps, and enable AI in it. (via extensions)

##  Actual Behavior
1. All ILogger traces are sent to AI including a lot from aspnet core runtime. Example Traces:
![image](https://user-images.githubusercontent.com/5232798/36328574-710d4b68-1317-11e8-8cdb-8e45048fb385.png)

## Expected Behavior
No traces from runtime should be collected by default. This is likely unwanted cost to users. 
This occurs only when starting the app from VS Debugging/WebApps.

The key issue here is the 'default' level of logs sent to AI. It seems like all logs above 'information' are sent to AI. From my experiment, it looks like every Request will have atleast 5 Info traces from asp.net runtime itself. This is most likely polluting user experience and costing money.

## Version Info 
SDK Version : 2.2 or 2.1
.NET Version :	.NET CORE 2.0 						
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) : VS
OS : Windows
Hosting Info (IIS/Azure WebApps/ etc) : Self-hosted locally

cijothomas:
Modify appsettings.json like below to control applicationinsights logging behaviour.
"Logging": {
    "IncludeScopes": false,

    "ApplicationInsights": {
      "LogLevel": {
        "Default": "None"
      }
    },
    "LogLevel": {
      "Default": "Warning"
    }
  }

cijothomas:
The default appsettings.json file for webapi in VS is:
{
  "Logging": {
    "IncludeScopes": false,
    "Debug": {
      "LogLevel": {
        "Default": "Warning"
      }
    },
    "Console": {
      "LogLevel": {
        "Default": "Warning"
      }
    }
  }
}
There is not default log level specified, so I believe for webapi anything >=Info is sent to AI.

cijothomas:
```
{
  "Logging": {
    "IncludeScopes": false,
    "LogLevel": {
      "Default": "Warning"
    }
  }
}
```
[...]

PR #853, opened 2019-03-20
Enable ILogger by default

cijothomas:
Fix Issue #854 
1. Take dependency of Microsoft.Extensions.Logging.ApplicationInsights for NETSTANDARD2.0
2. Enables ApplicationInsightsILogger Provider by default for Warning or above logs.
3. Removes the default DebugLogger - its irrelevant now, as Ilogger logs are now automatically captured.
4. Add FunctionalTests to validate ILogger integration.

- [ ] I ran Unit Tests locally.

- [ ] Changes in public surface reviewed
- [ ] Design discussion issue #
- [ ] CHANGELOG.md updated with one line description of the fix, and a link to the original issue.
- [ ] The PR will trigger build, unit tests, and functional tests automatically. If your PR was submitted from fork - mention one of committers to initiate the build for you.
	  If you want to to re-run the build/tests, the easiest way is to simply Close and Re-Open this same PR. (Just click 'close pull request' followed by 'open pull request' buttons at the bottom of the PR)

- Please follow [these] (https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/develop/Readme.md) instructions to build and test locally.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
@@ -22,8 +22,10 @@
     using Microsoft.AspNetCore.Http;
     using Microsoft.Extensions.Configuration;
     using Microsoft.Extensions.Configuration.Memory;
+    using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.DependencyInjection.Extensions;
     using Microsoft.Extensions.Options;
+    using Microsoft.Extensions.Logging;
 
     /// <summary>
     /// Extension methods for <see cref="IServiceCollection"/> that allow adding Application Insights services to application.
@@ -190,8 +192,6 @@
 
                 services.AddSingleton<TelemetryClient>();
 
-                services.AddSingleton<ApplicationInsightsDebugLogger, ApplicationInsightsDebugLogger>();
-
                 services
                     .TryAddSingleton<IConfigureOptions<ApplicationInsightsServiceOptions>,
                         DefaultApplicationInsightsServiceConfigureOptions>();
@@ -201,12 +201,28 @@
                 // that requires IOptions infrastructure to run and initialize
                 services.AddSingleton<IStartupFilter, ApplicationInsightsStartupFilter>();
 
-                services.AddSingleton<JavaScriptSnippet>();
-                services.AddSingleton<ApplicationInsightsLoggerEvents>();
+                services.AddSingleton<JavaScriptSnippet>();                
 
                 services.AddOptions();
                 services.AddSingleton<IOptions<TelemetryConfiguration>, TelemetryConfigurationOptions>();
                 services.AddSingleton<IConfigureOptions<TelemetryConfiguration>, TelemetryConfigurationOptionsSetup>();
+
+                // NetStandard2.0 has a package reference to Microsoft.Extensions.Logging.ApplicationInsights, and
+                // enables ApplicationInsightsLoggerProvider by default.
+                // This also adds LoggerMarker to the DI Container, which is used by LoggerFactory.AddApplicationInsights() method
+                // to detect ApplicationInsightsLoggerProvider as already added and back-off.
+                // Without this, there will be double logging if LoggerFactory.AddApplicationInsights() was already used.
+#if NETSTANDARD2_0
+                services.AddLogging(loggingBuilder =>
+                {
+                     loggingBuilder.AddApplicationInsights();
+
+                     // By default, all logs Warning or above is captured.
+                     // AddFilter is additive
+                     loggingBuilder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>("", LogLevel.Warning);
+                });                
+                services.TryAddSingleton<LoggerMarker>();
+#endif
             }
 
             return services;
@@ -406,8 +422,15 @@
 
         private static bool IsApplicationInsightsAdded(IServiceCollection services)
         {
-            // We treat ApplicationInsightsDebugLogger as a marker that AI services were added to service collection
-            return services.Any(service => service.ServiceType == typeof(ApplicationInsightsDebugLogger));
+            // We treat TelemetryClient as a marker that AI services were added to service collection
+            return services.Any(service => service.ServiceType == typeof(TelemetryClient));
         }
+    }
+
+    internal class LoggerMarker
+    {
+        // Serves no purpose other than acting as a Marker to indicate that ApplicationInsightsLoggerProvider was added.
+        // The presence of this marker is used by ILoggerFactory.AddApplicationInsights() method to detect
+        // ApplicationInsightsLoggerProvider as already added and back-off.
     }
 }
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
@@ -219,7 +219,7 @@
 
                      // By default, all logs Warning or above is captured.
                      // AddFilter is additive
-                     loggingBuilder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>("", LogLevel.Warning);
+                     loggingBuilder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>("Default", LogLevel.Warning);
                 });                
                 services.TryAddSingleton<LoggerMarker>();
 #endif
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
@@ -208,20 +208,30 @@
                 services.AddSingleton<IConfigureOptions<TelemetryConfiguration>, TelemetryConfigurationOptionsSetup>();
 
                 // NetStandard2.0 has a package reference to Microsoft.Extensions.Logging.ApplicationInsights, and
-                // enables ApplicationInsightsLoggerProvider by default.
-                // This also adds LoggerMarker to the DI Container, which is used by LoggerFactory.AddApplicationInsights() method
-                // to detect ApplicationInsightsLoggerProvider as already added and back-off.
-                // Without this, there will be double logging if LoggerFactory.AddApplicationInsights() was already used.
+                // enables ApplicationInsightsLoggerProvider by default.                
 #if NETSTANDARD2_0
                 services.AddLogging(loggingBuilder =>
                 {
                      loggingBuilder.AddApplicationInsights();
 
-                     // By default, all logs Warning or above is captured.
-                     // AddFilter is additive
-                     loggingBuilder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>("Default", LogLevel.Warning);
-                });                
-                services.TryAddSingleton<LoggerMarker>();
+                    // The default behavior is to capture only logs above Warning level from all categories.
+                    // This can achieved with this code level filter -> loggingBuilder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>("",LogLevel.Warning);
+                    // However, this will make it impossible to override this behavior from Configuration like below using appsettings.json:
+                    //"ApplicationInsights": {
+                    // "LogLevel": {
+                    // "": "Error"
+                    // }
+                    // },
+                    // The reason is as both rules will match the filter, the last one added wins.
+                    // To ensure that the default filter is in the beginning of filter rules, so that user override from Configuration will always win, 
+                    // we add code filter rule to the 0th position as below.
+
+                    loggingBuilder.Services.Configure<LoggerFilterOptions>
+                    (options => options.Rules.Insert(0,
+                        new LoggerFilterRule(
+                            "Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider", "",
+                            LogLevel.Warning, null)));
+                });                                
 #endif
             }
 
@@ -425,12 +435,5 @@
             // We treat TelemetryClient as a marker that AI services were added to service collection
             return services.Any(service => service.ServiceType == typeof(TelemetryClient));
         }
-    }
-
-    internal class LoggerMarker
-    {
-        // Serves no purpose other than acting as a Marker to indicate that ApplicationInsightsLoggerProvider was added.
-        // The presence of this marker is used by ILoggerFactory.AddApplicationInsights() method to detect
-        // ApplicationInsightsLoggerProvider as already added and back-off.
-    }
+    }   
 }
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
@@ -229,7 +229,7 @@
                     loggingBuilder.Services.Configure<LoggerFilterOptions>
                     (options => options.Rules.Insert(0,
                         new LoggerFilterRule(
-                            "Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider", "",
+                            "Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider", null,
                             LogLevel.Warning, null)));
                 });                                
 #endif
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
@@ -435,5 +435,5 @@
             // We treat TelemetryClient as a marker that AI services were added to service collection
             return services.Any(service => service.ServiceType == typeof(TelemetryClient));
         }
-    }   
+    }
 }
--- a/src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
@@ -66,31 +66,28 @@
             IServiceProvider serviceProvider,
             Func<string, LogLevel, bool> filter,
             Action loggerAddedCallback)
-        {
-            var loggerMarker = serviceProvider.GetService<LoggerMarker>();
-            if (loggerMarker == null)
+        {            
+            
+            var client = serviceProvider.GetService<TelemetryClient>();
+            var debugLoggerControl = serviceProvider.GetService<Microsoft.ApplicationInsights.AspNetCore.Logging.ApplicationInsightsLoggerEvents>();
+            var options = serviceProvider.GetService<IOptions<Microsoft.ApplicationInsights.AspNetCore.Logging.ApplicationInsightsLoggerOptions>>();
+
+            if (options == null)
             {
-                var client = serviceProvider.GetService<TelemetryClient>();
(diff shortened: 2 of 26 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
```diff
@@ -67,26 +67,30 @@ public static ILoggerFactory AddApplicationInsights(this ILoggerFactory factory,
             Func<string, LogLevel, bool> filter,
             Action loggerAddedCallback)
         {
-            var client = serviceProvider.GetService<TelemetryClient>();
-            var debugLoggerControl = serviceProvider.GetService<ApplicationInsightsLoggerEvents>();
-            var options = serviceProvider.GetService<IOptions<ApplicationInsightsLoggerOptions>>();
-
-            if (options == null)
+            var loggerMarker = serviceProvider.GetService<LoggerMarker>();
```
cijothomas: ConfigureServices() is guaranteed to be called before Configure() where this method is invoked, ensuring that LoggerMarker will be present by the code this code is executed.

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
```diff
@@ -95,10 +95,11 @@
...
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-  <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.0.0" />
-  <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.0.0" />
-  <PackageReference Include="Microsoft.Extensions.DiagnosticAdapter" Version="2.0.0" />
-  <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
+  <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.1.0" />
+  <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.1.0" />
+  <PackageReference Include="Microsoft.Extensions.DiagnosticAdapter" Version="2.1.0" />
+  <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.0" />
+  <PackageReference Include="Microsoft.Extensions.Logging.ApplicationInsights" Version="2.9.1" />
```
cijothomas: This also has Microsoft.ApplicationInsights dependency. So need to see how to avoid conflicts between this and the direct dependency for Microsoft.ApplicationInsights

cijothomas:
@SergeyKanzhelev @pakrym @RamjotSingh Please review when you get a chance! thanks in advance.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
RamjotSingh: Instead of doing this, maybe we can do IStartupFilter?

RamjotSingh: Essentially above in #If NetStandard2.0 we can put an additional #else and add a startup filter for starting up the old path.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerProvider.cs
```diff
@@ -7,7 +7,10 @@
     /// <summary>
     /// <see cref="ILoggerProvider"/> implementation that creates returns instances of <see cref="ApplicationInsightsLogger"/>
     /// </summary>
+#if !NETSTANDARD2_0
```
RamjotSingh: For clarity sake, put the whole class in #IF?

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -201,12 +201,28 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
...
+                // to detect ApplicationInsightsLoggerProvider as already added and back-off.
+                // Without this, there will be double logging if LoggerFactory.AddApplicationInsights() was already used.
+#if NETSTANDARD2_0
+                services.AddLogging(loggingBuilder =>
+                {
+                     loggingBuilder.AddApplicationInsights();
+
+                     // By default, all logs Warning or above is captured.
+                     // AddFilter is additive
+                     loggingBuilder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>("Default", LogLevel.Warning);
```
RamjotSingh: I wonder if we are messing with the ILogger infra here. This can already be done from the config which is read as part of ILogger setup. Hiding this statement this deep in SDK might be non-intuitive.

bradygaster: I'd be interested in @davidfowl's opinion here. Ideally, custom middleware authors would wire up all of the services requisite for their expected behavior. If someone uses `AddApplicationInsights`, for instance, we'd presume all trace, exception - everything, by default - goes to Application Insights, and that any service wire-ups required for that goal to be "easy" or even "invisible" to the developer.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerProvider.cs
```diff
@@ -7,7 +7,10 @@
...
+    // For NETSTANDARD2.0 We take dependency on Microsoft.Extensions.Logging.ApplicationInsights which has ApplicationInsightsProvider having the same ProviderAlias and don't want to clash with this ProviderAlias.
     [ProviderAlias("ApplicationInsights")]
+#endif
     internal class ApplicationInsightsLoggerProvider : ILoggerProvider
```
pakrym: Why do you keep this type instead of completely switching to Microsoft.Extensions.Logging.ApplicationInsights implementation?

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
```diff
@@ -86,19 +86,23 @@
...
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-  <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.0.0" />
-  <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.0.0" />
-  <PackageReference Include="Microsoft.Extensions.DiagnosticAdapter" Version="2.0.0" />
-  <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
+  <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.1.0" />
+  <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.1.0" />
+  <PackageReference Include="Microsoft.Extensions.DiagnosticAdapter" Version="2.1.0" />
+  <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.0" />
```
pakrym: Do you need this reference? I would be brought in by `Microsoft.Extensions.Logging.ApplicationInsights`

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
cijothomas: Since users updatng to this new SDK version has not explicitly made a decision to send ILogger logs to AI, i'd like to just sent Warning or above to AI. Without this, we could end up sending too much logs to AI. (negative surprises when billing comes). Since the runtime itself logs using ILogger, we have experienced user complaints about too much AI logs. (https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/603)

I'd open to all suggestions here. We want to balance between 'easy/auto capture of ILogger logs' and 'too much noise'.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
cijothomas: @RamjotSingh Is there any scenario broken with this? Or just a suggestion?

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerProvider.cs
cijothomas: We cannot remove this whole class - as that'd be a breaking change for AI customers, and we dont want that. (except when we change major from to 3.XX)..

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
RamjotSingh: Probably not broken (not sure). But this code looked like it was abusing a DI construct. Hence the suggestion.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
cijothomas: By default, only "Warning" or above LogLevel logs are sent to ApplicationInsightsProvider for all categories. But this can be easily configured by user either in code or using appsettings.json.  
I added more description here: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/854

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerProvider.cs
cijothomas: We cannot remove this whole class - as that'd be a breaking change for AI customers, and we don't want that. (except when we change major from to 3.XX).

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
cijothomas: right.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
cijothomas: This need revisit. By doing this, i am essentially breaking behavior already - i.e users already using logerFactory.AddApplicationInsights will see that their Filters are not respected, as this PR now favors the new Provider.

Need to do the inverse somehow - i.e for existing loggerFactory.AddApplicationInsights users, the new Provider should be disabled.

cijothomas: We don't need to detect this actually. Since the old ILoggerFactory method is marked obsolete, we expect customers to stop using it. I will improve the Obsolete warning to be more descriptive and link to official docs.

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
cijothomas: This is not a new issue.Same with taking dependency of Perf/Dep packages. We now need to ensure loggingis also updated with latest base sdk.

src/Microsoft.ApplicationInsights.AspNetCore/Logging/Implementation/ApplicationInsightsLoggerFactoryExtensions.cs
cijothomas: closing this as not longer applicable.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -201,12 +201,38 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
...
+                    // }
+                    // },
+                    // The reason is as both rules will match the filter, the last one added wins.
+                    // To ensure that the default filter is in the beginning of filter rules, so that user override from Configuration will always win, 
+                    // we add code filter rule to the 0th position as below.
+
+                    loggingBuilder.Services.Configure<LoggerFilterOptions>
+                    (options => options.Rules.Insert(0,
+                        new LoggerFilterRule(
+                            "Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider", "",
```
pakrym: The category is usually set to null for "all categories" case.

cijothomas: fixed.

cijothomas:
@SergeyKanzhelev can you review?

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -406,8 +432,8 @@ public static IServiceCollection AddApplicationInsightsTelemetryProcessor<T>(thi
 
         private static bool IsApplicationInsightsAdded(IServiceCollection services)
```
SergeyKanzhelev: what is this method used for? Why the change?

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
```diff
@@ -86,17 +86,20 @@
     <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.10.0-beta2" />
     <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer" Version="2.10.0-beta2" />
     <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel" Version="2.10.0-beta2" />
-    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="1.0.2" />
-    <PackageReference Include="Microsoft.Extensions.Configuration" Version="1.0.2" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="1.0.2" />
```
SergeyKanzhelev: will `Microsoft.AspNetCore.Hosting.Abstractions` be enough?

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
```diff
@@ -86,17 +86,20 @@
...
-    <PackageReference Include="Microsoft.Extensions.Configuration" Version="1.0.2" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="1.0.2" />           
+    <PackageReference Include="System.Text.Encodings.Web" Version="4.3.1" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' != 'netstandard2.0' ">
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="1.0.2" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.0.2" />
-    <PackageReference Include="System.Text.Encodings.Web" Version="4.3.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="1.0.2" />
```
RamjotSingh: Same as above, can .abstractions be enough?

RamjotSingh: Actually this should already be getting imported coz of Microsoft.Extensions.Configuration.Json

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
```diff
@@ -86,17 +86,20 @@
...
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-  <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.0.0" />
-  <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.0.0" />
-  <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
+  <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.1.0" />
+  <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.1.0" />
```
RamjotSingh: This should also be coming from Microsoft.Extensions.Logging.AI so you can skip it here.

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
cijothomas: we use this to detect if AI services are already added to DI to prevent double injection if customer calls both AddApplicationInsightsTelemetry() and UseApplicationInsights(). We previously relied on presence of 'ApplicationInsightsDebugLogger' to detect this. With that no longer needed (since ilogger is automatically enabled), we are now relying on presense of telemetryclient to detect if AI is already added.

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
cijothomas: Likely yes, but i don't want to make un-related changes to this PR. Will revisit all package references in a separate PR.

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
cijothomas: Likely yes, but i don't want to make un-related changes to this PR. Will revisit all package references in a separate PR.

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
cijothomas: Likely yes, but i don't want to make un-related changes to this PR. Will revisit all package references in a separate PR.

src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
davidfowl: We should do that because a bunch of these dependencies look suspect

src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -201,12 +201,38 @@ public static IServiceCollection AddApplicationInsightsTelemetry(this IServiceCo
...
-                services.AddSingleton<ApplicationInsightsLoggerEvents>();
+                services.AddSingleton<JavaScriptSnippet>();                
 
                 services.AddOptions();
                 services.AddSingleton<IOptions<TelemetryConfiguration>, TelemetryConfigurationOptions>();
                 services.AddSingleton<IConfigureOptions<TelemetryConfiguration>, TelemetryConfigurationOptionsSetup>();
+
+                // NetStandard2.0 has a package reference to Microsoft.Extensions.Logging.ApplicationInsights, and
+                // enables ApplicationInsightsLoggerProvider by default.                
+#if NETSTANDARD2_0
```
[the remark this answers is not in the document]
pharring: NetFX will always "win" over netstandard2.0

Issue #862, opened 2019-04-08
HttpRequestExtensions.GetUri throws Invalid URI when multiple Host headers are present


saikatguha:
If you are reporting bug/issue, please provide detailed Repro instructions.

## Repro Steps
1. Send HTTP request with multiple Host headers.
2. Observe 500 Internal Error

Reason is that `request.Host.ToString()` returns _comma-separated list_ of all the Host header values. If only one Host header is present, all is fine. However, if multiple Host headers are present, it results in a malformed URI being created in [line 38](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/c9b04f5d2fbf97a0687ac45f33f26c5b6d705603/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/HttpRequestExtensions.cs#L38) of the form `https://host1,host2/path`

##  Actual Behavior
```
System.UriFormatException: Invalid URI: The hostname could not be parsed.
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   at System.Uri..ctor(String uriString)
   at Microsoft.ApplicationInsights.AspNetCore.Extensions.HttpRequestExtensions.GetUri(HttpRequest request)
   at Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners.HostingDiagnosticListener.EndRequest(HttpContext httpContext, Int64 timestamp)
 ...
```

## Expected Behavior

Request completes as normal.

## Version Info 
SDK Version : 2.6.1
.NET Version : core
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) :
OS :
Hosting Info (IIS/Azure WebApps/ etc) : Azure WebApps

cijothomas:
@Dmitry-Matveev You are already working on get URI right? Can you take a look at this ?

Dmitry-Matveev:
@saikatguha, yep, I concur, the previous and the existing implementations are both subject to this issue. We now take `Value` if it exists, but `Value` can still be malformed for URI purposes, I guess.

AI is not supposed to fail the app under any circumstance, though - did it fail your request and service?
@cijothomas , do we still have try-catch all rule across SDKs?

saikatguha:
@Dmitry-Matveev yes, all requests with multiple host headers fail before returning response to the client. If I remove AI logger, requests succeed.

cijothomas:
[...]

PR #864, opened 2019-04-11
Performance fixes, wave 1.

Dmitry-Matveev:
- Swtich from Diagnostics adapters to listeners from @lmolkova (was already merged into develop, this commit is therefore ignored by PR);
- Switch from `Initialize()` to `InitializeInsturmenationKey()` to prevent second round of Telemetry Initializers
- `GetUri` updates, 2x gain ([this bug](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/862) is still an issue, though)
- `SetHeaderKeyValue` updates, 10x gain
- Use deprecated `Properties` for a while again to avoid second concurrent dictionary initialization for `GlobalProperties`.
- Test updates, including awesome contribution from @lmolkova to fix W3C correlation the aforementioned changes broke

- [x] CHANGELOG.md updated with one line description of the fix, and a link to the original issue.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HeadersUtilities.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HeadersUtilities.cs
@@ -5,6 +5,7 @@ namespace Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners
     using System.Linq;
     using System.Text.RegularExpressions;
     using Microsoft.ApplicationInsights.Common;
+    using Microsoft.Extensions.Primitives;
 
     /// <summary>
     /// Generic functions that can be used to get and set Http headers.
@@ -20,7 +21,7 @@ namespace Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners
         public static string GetHeaderKeyValue(IEnumerable<string> headerValues, string keyName)
         {
             if (headerValues != null)
-            {
+            {                
                 foreach (string keyNameValue in headerValues)
                 {
                     string[] keyNameValueParts = keyNameValue.Trim().Split('=');
@@ -54,6 +55,27 @@ namespace Microsoft.ApplicationInsights.AspNetCore.DiagnosticListeners
                 : headerValues
                     .Where(headerValue => !HeaderMatchesKey(headerValue, keyName))
                     .Concat(newHeaderKeyValue);
+        }
+
+        internal static StringValues SetHeaderKeyValue(string[] currentHeaders, string key, string value)
+        {
+            if (currentHeaders != null)
+            {
+                for (int index = 0; index < currentHeaders.Length; index++)
+                {
+                    if (HeaderMatchesKey(currentHeaders[index], key))
+                    {
+                        currentHeaders[index] = string.Concat(key, "=", value);
+                        return currentHeaders;
+                    }
+                }
+
+                return StringValues.Concat(currentHeaders, string.Concat(key, "=", value));                
+            }
+            else
+            {
+                return string.Concat(key, "=", value);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HeadersUtilities.cs
```diff
@@ -38,22 +39,33 @@ public static string GetHeaderKeyValue(IEnumerable<string> headerValues, string
...
-                : headerValues
-                    .Where(headerValue => !HeaderMatchesKey(headerValue, keyName))
-                    .Concat(newHeaderKeyValue);
+            if (currentHeaders != null)
+            {
+                for (int index = 0; index < currentHeaders.Length; index++)
+                {
+                    if (HeaderMatchesKey(currentHeaders[index], key))
+                    {
+                        currentHeaders[index] = string.Concat(key, "=", value);
```
Dmitry-Matveev: Note: Tried making it a variable, IL adds an extra allocation for that, so kept in three places on purpose.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AspNetCoreEnvironmentTelemetryInitializer.cs
```diff
@@ -23,10 +23,10 @@ public AspNetCoreEnvironmentTelemetryInitializer(IHostingEnvironment environment
 
         /// <inheritdoc />
         public void Initialize(ITelemetry telemetry)
-        {
-            if (environment != null && !telemetry.Context.GlobalProperties.ContainsKey(AspNetCoreEnvironmentPropertyName))
+        {   
+            if (environment != null && !telemetry.Context.Properties.ContainsKey(AspNetCoreEnvironmentPropertyName))
             {
-                telemetry.Context.GlobalProperties.Add(AspNetCoreEnvironmentPropertyName, environment.EnvironmentName);
+                telemetry.Context.Properties.Add(AspNetCoreEnvironmentPropertyName, environment.EnvironmentName);
```
cijothomas: If we can avoid touching properties at all in default scenarios, then that'll also save the perf hit from instantiating the Properties Conc.Dictionary.

Can you try if changes similar to this(https://github.com/Microsoft/ApplicationInsights-dotnet/pull/929/files) can be applied he as well? We'll need to modify all telemetry types to have this special internal field.

https://github.com/Microsoft/ApplicationInsights-dotnet/pull/929/files

Dmitry-Matveev: Actually, don't we stamp each request with Pre-Aggregation's "_MS.AggregatedByMetricExtractors" or similar? Which dictionary does it use - we should just use that one since it's initialized for each request anyway.

src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HeadersUtilities.cs
pharring: That sounds suspicious. Are you counting locals (stack slots) from IL? Those aren't the kind of allocations (GC heap) to be worried about.

Dmitry-Matveev: Yep. ldloc.1 and similar. Memory hit here is fine, i'd like to avoid extra instructions if possible :)
Not to the point of one allocation of course, but no allocation simply sounded better than "an allocation" upon review of IL here. Not a problem, can move to a variable if you feel it's absolutely necessary.

pharring: IL instruction count isn't really that important. Sure, fewer is generally better (you can fit more on a cache line) but, of course, it's the generated machine code that really matters. An IL "slot" is required for local variables (and args) and they'll usually map to a register or a BP-relative memory location. We might call the latter "stack allocation", but it's nowhere near the same as a GC allocation. In this case, I'd expect that the total IL instruction count for this method when hoisting the expression to a local variable (common subexpression elimination) will be less overall -- because of the saved duplicate code. I doubt the C# compiler will do the hoisting itself -- and if it did, it'd use a stack slot.

This is all a long explanation for why you shouldn't worry about IL slots for perf.

For this case, I don't mind how you write it. I just wanted to explain why "IL adds an extra allocation" raised my Spock-like eyebrow and, hopefully, educate the team :-)

Dmitry-Matveev: Cool! Great to know!

PR #888, opened 2019-05-24
sdl fix System.Net.Security and System.Net.Http

MS-TimothyMothra:
taking an explicit dependency on System.Net.Http will override any implicit dependencies.

Confirmed that this fixes the security scan.

I'm trying to copy @lmolkova's solution from here: https://github.com/microsoft/ApplicationInsights-dotnet/pull/1147

@lmolkova expressed some concerns about this in Web repo. Lets not merge until we resolve all concerns and agree this is the correct approach.


lmolkova:
I think this is not needed, as base sdk will bring required version of system.net.http in 2.10-beta1

MS-TimothyMothra:
I created a build of latest Base SDK and verified that it fixes the issues in Logging SDK. It does! :)
I think we can close this issue. 
Thanks @lmolkova

PR #930, opened 2019-07-13
Log ApplicationInsightsStartupFilter exceptions

davidfowl:
- This will end up in application logs instead of logging solely into the event source

cc @cijothomas


davidfowl:
@cijothomas it might be worth also doing another pass looking for places we log via AspNetCoreEventSource and log via an ILogger as well.

cijothomas:
@davidfowl  adding an issue to track add more ILogger logs in more places as feasible.
https://github.com/microsoft/ApplicationInsights-aspnetcore/issues/935

PR #971, opened 2019-09-06
Do not set sampledOut flag if activity is recorded and parentId fixes

lmolkova:
If Activity is recorded, we are going to sample in proactively (i.e. give more weight for these items, see https://github.com/Microsoft/ApplicationInsights-dotnet/pull/1200)

When we proactively sample features out, we should check Activity.Recorded first and only sample out those which are not recorded.

Also, I noticed parent -id on requests is not set correctly in W3C format. fixing that.

### DIFF
--- a/test/Microsoft.ApplicationInsights.AspNetCore.Tests/HostingDiagnosticListenerTest.cs
+++ b/test/Microsoft.ApplicationInsights.AspNetCore.Tests/HostingDiagnosticListenerTest.cs
@@ -459,7 +459,8 @@
         [InlineData(AspNetCoreMajorVersion.One, false)]
         [InlineData(AspNetCoreMajorVersion.Two, false)]
         [InlineData(AspNetCoreMajorVersion.Three, false)]
-        public void RequestWithW3CTraceParentCreateNewActivityAndPopulateRequestTelemetry(AspNetCoreMajorVersion aspNetCoreMajorVersion, bool IsW3C)
+        public void RequestWithW3CTraceParentCreateNewActivityAndPopulateRequestTelemetry(AspNetCoreMajorVersion aspNetCoreMajorVersion, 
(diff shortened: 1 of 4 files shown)


test/Microsoft.ApplicationInsights.AspNetCore.Tests/HostingDiagnosticListenerTest.cs
```diff
@@ -1014,7 +1021,35 @@ public void RequestTelemetryIsProactivelySampledOutIfFeatureFlagIsOn(AspNetCoreM
...
         [InlineData(AspNetCoreMajorVersion.Three)]
-        public void RequestTelemetryIsNotProactivelySampledOutIfFeatureFlasIfOff(AspNetCoreMajorVersion aspNetCoreMajorVersion)
+        public void RequestTelemetryIsProactivelySampledInIfFeatureFlagIsOnButActivityIsRecorded(AspNetCoreMajorVersion aspNetCoreMajorVersion)
+        {
+            TelemetryConfiguration config = TelemetryConfiguration.CreateDefault();
+            config.ExperimentalFeatures.Add("proactiveSampling");
+            config.SetLastObservedSamplingPercentage(SamplingTelemetryItemTypes.Request, 0);
+
+            HttpContext context = CreateContext(HttpRequestScheme, HttpRequestHost, "/Test", method: "POST");
+            var traceParent = "00-4e3083444c10254ba40513c7316332eb-e2a5f830c0ee2c46-01";
```
cijothomas: perhaps a test with -00 for Recorded flag?

lmolkova: this one tests -01 request with sampled out flag set to 100% of items, so this is intentionally -01

PR #992, opened 2019-09-24
SDK Connection Strings (Net Core support)

TimothyMothra:
https://github.com/microsoft/ApplicationInsights-dotnet/issues/1221

**TODO**
- Need to discuss Javascript Snippit with SDK team.

**Changes**
- Extensions and ServiceOptions

### DIFF
--- a/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
+++ b/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
@@ -21,7 +21,7 @@
             this.EnablePerformanceCounterCollectionModule = true;
             this.EnableQuickPulseMetricStream = true;
             this.EnableAdaptiveSampling = true;
-            this.EnableDebugLogger = true;            
+            this.EnableDebugLogger = true;
             this.EnableHeartbeat = true;
             this.AddAutoCollectedMetricExtractor = true;
 #if AI_ASPNETCORE_WEB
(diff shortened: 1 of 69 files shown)


test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/ApplicationInsightsExtensionsTests.cs
```diff
@@ -201,15 +215,38 @@ public static void RegistersTelemetryConfigurationFactoryMethodThatReadsInstrume
...
+            public static void AddApplicationInsightsTelemetry_ReadsConnectionString_FromEnvironment()
+            {
+                var services = ApplicationInsightsExtensionsTests.GetServiceCollectionWithContextAccessor();
+                Environment.SetEnvironmentVariable(ConnectionStringEnvironmentVariable, TestConnectionString);
+                try
+                {
+                    services.AddApplicationInsightsTelemetry();
+                    IServiceProvider serviceProvider = services.BuildServiceProvider();
+                    var telemetryConfiguration = serviceProvider.GetTelemetryConfiguration();
+                    Assert.Equal(TestConnectionString, telemetryConfiguration.ConnectionString);
```
cijothomas: validate that config.ikey, endpoint etc are also set from con string.

test/Microsoft.ApplicationInsights.AspNetCore.Tests/Extensions/ApplicationInsightsExtensionsTests.cs
```diff
@@ -201,15 +215,38 @@ public static void RegistersTelemetryConfigurationFactoryMethodThatReadsInstrume
...
             /// <summary>
             /// Validates that while using services.AddApplicationInsightsTelemetry(); ikey is read from
             /// Environment
             /// </summary>
             [Fact]
             public static void AddApplicationInsightsTelemetryReadsInstrumentationKeyFromEnvironment()
             {
-                var services = ApplicationInsightsExtensionsTests.GetServiceCollectionWithContextAccessor();                
-                Environment.SetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY", TestInstrumentationKey);                
+                var services = ApplicationInsightsExtensionsTests.GetServiceCollectionWithContextAccessor();
```
cijothomas: lets add one test with both ikey,endpoint in appsettings.json AND connectionstring also on appsettings.json, Validate that connectionstring Wins.

src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -65,6 +65,11 @@ public void Configure(TelemetryConfiguration configuration)
                     configuration.InstrumentationKey = this.applicationInsightsServiceOptions.InstrumentationKey;
                 }
 
+                if (this.applicationInsightsServiceOptions.ConnectionString != null)
```
cijothomas: we read endpoint address further down this class. That should be moved before ConString. Else it'll override connectionstring's endpoint address.

src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -117,6 +117,12 @@ public void Configure(TelemetryConfiguration configuration)
                     configuration.TelemetryChannel.EndpointAddress = this.applicationInsightsServiceOptions.EndpointAddress;
                 }
 
+                // Need to set connection string before calling Initialize() on the Modules and Processors.
+                if (this.applicationInsightsServiceOptions.ConnectionString != null)
+                {
+                    configuration.ConnectionString = this.applicationInsightsServiceOptions.ConnectionString;
```
cijothomas: will this overwrite the channel.endpoint as well?

src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
```diff
@@ -164,6 +169,7 @@ internal void CopyPropertiesTo(ApplicationInsightsServiceOptions target)
                 target.InstrumentationKey = this.InstrumentationKey;
             }
 
+            target.ConnectionString = this.ConnectionString;
```
cijothomas: do you need to check constring for null or empty?

TimothyMothra:
I didn't see your comments until after I clicked merge. I'm reviewing now.

src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
TimothyMothra: this isn't necessary. it won't affect the behavior to copy the properties as is from one ServiceOptions to the other

src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs
TimothyMothra: Yes, which is why I put it last.

Issue #373, opened 2017-03-29
Synchronize access to HttpContext


SergeyKanzhelev:
`HttpContext` class is not thread safe. So accessing it from telemetry initializers on the different threads have to be synchronized. Today enabling Application Insights on highly async web site with many tasks running in parallel may cause failures or hangs.

```
000000aa`844d7fe0 00007ffc`97b8d704 mscorlib_ni!System.Collections.Generic.Dictionary`2[[System.__Canon, mscorlib],[System.__Canon, mscorlib]].FindEntry(System.__Canon)+0x10c
000000aa`844d8050 00007ffc`39d0cb71 mscorlib_ni!System.Collections.Generic.Dictionary`2[[System.__Canon, mscorlib],[System.__Canon, mscorlib]].TryGetValue(System.__Canon, System.__Canon ByRef)+0x14
000000aa`844d8090 00007ffc`39d0cae6 Microsoft_AspNetCore_Http_Features!Microsoft.AspNetCore.Http.Features.FeatureCollection.get_Item(System.Type)+0x41
000000aa`844d80d0 00007ffc`39f1afd2 Microsoft_AspNetCore_Http_Features!Microsoft.AspNetCore.Http.Features.FeatureCollection.Get[[System.__Canon, mscorlib]]()+0x36
000000aa`844d8110 00007ffc`39f1ae38 Microsoft_AspNetCore_Http_Features!Microsoft.AspNetCore.Http.Features.FeatureReferences`1[[System.__Canon, mscorlib]].UpdateCached[[System.__Canon, mscorlib],[System.__Cano
non>, Int32, Boolean)+0x72
000000aa`844d8180 00007ffc`39f1a3f6 Microsoft_AspNetCore_Http!Microsoft.AspNetCore.Http.Features.RequestCookiesFeature.get_Cookies()+0xc8
```

SergeyKanzhelev:
cc: @pakrym @davidfowl

SergeyKanzhelev:
**TL;DR;** We have a few ideas how to make Application Insights safer, but to really fix the issue we need to have a guidance and/or API in ASP.NET for locking the HttpContext object access.

So the question is how to fix the issue without running into the problem of parallel access to the `HttpContext` between Application Insights, other tracing/telemetry providers and customer code. Basically we have three scenarios for Application Insights:

1. Two telemetry items related to the same `HttpContext` being traced in parallel. Here we have two options:
[...]

PR #1006, opened 2019-10-14
Add new RoleName initializer for Azure Web App to accurately reflect slot swaps

cijothomas:
Fix Issue # https://github.com/microsoft/ApplicationInsights-dotnet-server/issues/1207

There'll be similar fix for Asp.Net Apps in https://github.com/microsoft/ApplicationInsights-dotnet-server repo as well.

### DIFF
--- a/src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
+++ b/src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
@@ -1,0 +1,149 @@
+﻿namespace Microsoft.ApplicationInsights.AspNetCore.TelemetryInitializers
+{
+    using System;
+    using Microsoft.ApplicationInsights.AspNetCore.Extensibility.Implementation.Tracing;
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+    using Microsoft.AspNetCore.Http;
+
+    /// <summary>
+    /// A telemetry initializer that will gather Azure Web App Role Environment context information to
+    /// populate TelemetryContext.Cloud.RoleName
+    /// This uses the http header "WAS-DEFAULT-HOSTNAME" to update role name, if available.
+    /// Otherwise role name is populated from "WEBSITE_HOSTNAME" environment variable.
+    /// </summary>
+    /// <remarks>
+    /// The RoleName is expected to contain the host name + slot name, but will be same across all instances of
+    /// a single App Service.
+    /// Populating RoleName from HOSTNAME environment variable will cause RoleName to be incorrect when a slot swap occurs in AppService.
+    /// The most accurate way to determine the RoleName is to rely on the header WAS-DEFAULT-HOSTNAME, as its
+    /// populated from App service front end on every request. Slot swaps are instantly reflected in this header.
+    /// </remarks>
+    public class AzureAppServiceRoleNameFromHostNameHeaderInitializer : ITelemetryInitializer
+    {
+        private const string WebAppHostNameHeaderName = "WAS-DEFAULT-HOSTNAME";
+        private const string WebAppHostNameEnvironmentVariable = "WEBSITE_HOSTNAME";
+        private readonly IHttpContextAccessor httpContextAccessor;
+        private string roleName;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AzureAppServiceRoleNameFromHostNameHeaderInitializer" /> class.
+        /// </summary>
+        /// <param name="httpContextAccessor">Accessor to provide HttpContext if available.</param>
+        public AzureAppServiceRoleNameFromHostNameHeaderInitializer(IHttpContextAccessor httpContextAccessor)
+        {
+            this.httpContextAccessor = httpContextAccessor ?? throw new ArgumentNullException(nameof(httpContextAccessor));
+
+            try
+            {
+                var result = Environment.GetEnvironmentVariable(WebAppHostNameEnvironmentVariable);
+
+                if (!string.IsNullOrEmpty(result) && result.EndsWith(this.WebAppSuffix, StringComparison.OrdinalIgnoreCase))
+                {
+                    result = result.Substring(0, result.Length - this.WebAppSuffix.Length);
+                }
+
+                this.roleName = result;
+            }
+            catch (Exception ex)
+            {
+                AspNetCoreEventSource.Instance.LogAzureAppServiceRoleNameFromHostNameHeaderInitializerWarning(ex.ToInvariantString());
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets suffix of website name. This must be changed when running in non public Azure region.
+        /// Default value (Public Cloud):  ".azurewebsites.net"
+        /// For US Gov Cloud: ".azurewebsites.us"
+        /// For Azure Germany: ".azurewebsites.de".
+        /// </summary>
+        public string WebAppSuffix { get; set; } = ".azurewebsites.net";
+
+        /// <summary>
+        /// Populates RoleName from the request telemetry associated with the http context.
+        /// If RoleName is empty on the request telemetry, it'll be updated as well so that other telemetry
+        /// belonging to the same requests gets it from request telemetry, without having to parse headers again.
+        /// </summary>
+        /// <remarks>
+        /// RoleName is attempted from every incoming request as opposed to doing this periodically. This is
+        /// done to ensure every request (and associated telemetry) gets the correct RoleName during slot swap.
+        /// </remarks>
+        /// <param name="telemetry">The telemetry item for which RoleName is to be set.</param>
+        public void Initialize(ITelemetry telemetry)
+        {
+            try
+            {
+                if (!string.IsNullOrEmpty(telemetry.Context.Cloud.RoleName))
+                {
+                    // RoleName is already populated.
+                    return;
+                }
+
+                string roleName = string.Empty;
+                var context = this.httpContextAccessor.HttpContext;
+
+                if (context != null)
+                {
+                    lock (context)
+                    {
+                        var request = context.Features.Get<RequestTelemetry>();
+
+                        if (request != null)
+                        {
+                            if (string.IsNullOrEmpty(request.Context.Cloud.RoleName))
+                            {
+                                roleName = this.GetRoleNameFromHeader(context);
+                                if (!string.IsNullOrEmpty(roleName))
+                                {
+                                    request.Context.Cloud.RoleName = roleName;
+                                }
+                            }
+                            else
+                            {
+                                roleName = request.Context.Cloud.RoleName;
+                            }
+                        }
+                        else
+                        {
+                            roleName = this.GetRoleNameFromHeader(context);
+                        }
+                    }
+                }
+
+                if (string.IsNullOrEmpty(roleName))
+                {
+                    // Fallback to value from ENV variable.
+                    roleName = this.roleName;
+                }
+
+                telemetry.Context.Cloud.RoleName = roleName;
+            }
+            catch (Exception ex)
+            {
+                AspNetCoreEventSource.Instance.LogAzureAppServiceRoleNameFromHostNameHeaderInitializerWarning(ex.ToInvariantString());
+            }
+        }
+
+        private string GetRoleNameFromHeader(HttpContext context)
+        {
+            string roleName = string.Empty;
+            if (context.Request?.Headers != null)
+            {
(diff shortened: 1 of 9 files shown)


src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -0,0 +1,149 @@
...
+                    // RoleName is already populated.
+                    return;
+                }
+
+                string roleName = string.Empty;
+                var context = this.httpContextAccessor.HttpContext;
+
+                if (context != null)
+                {
+                    lock (context)
```
TimothyMothra: can you add a comment.. I don't understand why is the context being locked here?

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -0,0 +1,149 @@
...
+                    {
+                        var request = context.Features.Get<RequestTelemetry>();
+
+                        if (request != null)
+                        {
+                            if (string.IsNullOrEmpty(request.Context.Cloud.RoleName))
+                            {
+                                roleName = this.GetRoleNameFromHeader(context);
+                                if (!string.IsNullOrEmpty(roleName))
```
TimothyMothra: I think this could be simplified by changing it to a `if (this.TryGetRoleNameFromHeader(context, out roleName)` so you won't need to repeat the IsNullOrEmpty check.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
cijothomas: One of the oldest issues we have: https://github.com/microsoft/ApplicationInsights-aspnetcore/issues/373
(if you check all other initializers in this repo, you'll see same lock)

cijothomas:
test failures are being fixed now.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -0,0 +1,149 @@
...
+                string roleName = string.Empty;
+                var context = this.httpContextAccessor.HttpContext;
+
+                if (context != null)
+                {
+                    lock (context)
+                    {
+                        var request = context.Features.Get<RequestTelemetry>();
+
+                        if (request != null)
```
pharring: I know this might sound odd, but is it possible that `request` is the same as `this` here? If so, we can short-circuit.

src/Microsoft.ApplicationInsights.WorkerService/ApplicationInsightsExtensions.cs
```diff
@@ -89,6 +90,7 @@ public static IServiceCollection AddApplicationInsightsTelemetryWorkerService(th
             {
                 if (!IsApplicationInsightsAdded(services))
                 {
+                    services.AddSingleton<ITelemetryInitializer, AzureWebAppRoleEnvironmentTelemetryInitializer>();
```
pharring: You're adding something that depends on HTTP (IHttpContextAcessor) to the WorkerService nuget...?
That seems misplaced. Can you explain?

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets suffix of website name. This must be changed when running in non public Azure region.
+        /// Default value (Public Cloud):  ".azurewebsites.net"
+        /// For US Gov Cloud: ".azurewebsites.us"
+        /// For Azure Germany: ".azurewebsites.de".
+        /// </summary>
+        public string WebAppSuffix { get; set; } = ".azurewebsites.net";
```
pharring: For .NET Framework I presume you could configure from ApplicationInsights.config, but how do you configure it for .NET Core?

cijothomas: Following should work for Asp.net core:
services.AddSingleton<ITelemetryInitializer>(new MyTelemetryInitializer() { fieldName = "myfieldName" });
(Agree its not mentioned in the doc  https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core#adding-telemetryinitializers, but i can add there)

There exists no other **easy** alternate to add a telemetry initializer with fields.

cijothomas: This is not a new problem - we already have telemetryinitializers which has public properties.

src/Microsoft.ApplicationInsights.WorkerService/ApplicationInsightsExtensions.cs
cijothomas: `AzureAppServiceRoleNameFromHostNameHeaderInitializer` is the name of the new initializer which relies on httpcontext.`AzureWebAppRoleEnvironmentTelemetryInitializer` is the old implementation relying purely on ENV variables, and it is continued to be used in WorkerService.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
lmolkova: you could easily avoid lock and access to the http context if you parsed the header in the request tracking module and populated it somewhere. Initializer would took the value from there.
This avoids accessing HttpContext on each telemetry and lock on each telemetry.

Please also consider side effects like one we learned yesterday about (https://github.com/aspnet/AspNetCore/issues/14975) that availability and integrity of HttpContext at the moment when telemetry is tracked is not guaranteed and overall, HttpContext is not recommended API to use.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
cijothomas: yes its possible and will occur for every auto collected Request telemetry. But I am not sure if we gain anything by checking if they are same. Can you help understand where can we optimize?

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
cijothomas: Yes its some thing which should be explored and implemented. There are several other initializers which can benefit from this optimization. (almost all initializers derived from TelemetryInitializerBase).

However, I do not think it's blocking this PR. It should be pursued as a separate improvement to benefit all initializers, including this.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
cijothomas: https://github.com/MicrosoftDocs/azure-docs/issues/40819

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
lmolkova: Reality is that we introduce yet another contention (because perf is not good already) that we should fix separately, but we likely never will. At least this work is not planned.  It seems doing a more performant thing is not that complicated, so why delay it for unknown amount of time?

cijothomas: I'd prefer to not delay this fix for another milestone (after being pushed several times already). This is a constant source of support tickets with no end in sight.

Please post a new issue in github about the proposed way to avoid locking, I will take it for all the initializers. 

If perf becomes a concern due to this particular initializer, it can removed and replaced with the old initializer. (by user)

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -0,0 +1,164 @@
...
+        private const string WebAppHostNameEnvironmentVariable = "WEBSITE_HOSTNAME";
+        private readonly IHttpContextAccessor httpContextAccessor;
+        private string roleName;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AzureAppServiceRoleNameFromHostNameHeaderInitializer" /> class.
+        /// </summary>
+        /// <param name="httpContextAccessor">Accessor to provide HttpContext if available.</param>
+        public AzureAppServiceRoleNameFromHostNameHeaderInitializer(IHttpContextAccessor httpContextAccessor)
+            : this(httpContextAccessor, ".azurewebsites.net")
```
reyang: In Mooncake, this would be `*.chinacloudsites.cn` - how do we handle these region specific things systematically?

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets suffix of website name. This must be changed when running in non public Azure region.
+        /// Default value (Public Cloud):  ".azurewebsites.net"
+        /// For US Gov Cloud: ".azurewebsites.us"
+        /// For Azure Germany: ".azurewebsites.de".
+        /// </summary>
+        public string WebAppSuffix { get; set; }
```
reyang: Do we ask the user to configure this?

cijothomas: The previous initializer did not allow configuring this, so users were left with no choice.
Now its made configurable, and defaults to public cloud. User has to explicitly change this if they are running in non-public regions. SDK has no automatic mechanism to detect this.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
cijothomas: Today SDK does not have any logic to deal with this. Its entirely up to the user to override these.

src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
reyang: This seems like opening another door of support tickets and a TL;DR document. Is there a way to detect this automatically?

cijothomas: I haven't explored that option. (i.e SDK detecting which cloud we are running..) We could do simple things like have a list of known suffixes for all known suffixes, or more complex by leveraging something from connection strings. I haven't explored the merits/demerits of each.
 
The support problem already exists and is not introduced by this change. This PR at least offers a workaround, where previously no workaround existed. In my opinion, this is an improvement to the current state.

reyang: I see. Thanks for the context.

cijothomas:
Will post a followup issue to address suggestion from this PR which were not addressed due to time constraints.

PR #1015, opened 2019-10-21
Reorg: move files from root to NETCORE directory

TimothyMothra:
https://github.com/microsoft/ApplicationInsights-dotnet/issues/1214

Step 1. Reorganize individual repos. This is to ensure that there are no conflicting directories or file names and to guarantee no conflicts when migrating source. This will require some minor changes to existing build definitions.

**Changes**

Update Directory
- EnlistmentRoot stays in root directory
- Add SourceRoot


.vs/ApplicationInsights.AspNetCore/config/applicationhost.config
```diff
@@ -0,0 +1,990 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
```
cijothomas: Whats the content of .vs folder?

cijothomas (changes requested):
do we need the .vs folder? 
That aside, the only change is move entire repo inside NETCORE folder, right?

.vs/ApplicationInsights.AspNetCore/config/applicationhost.config
TimothyMothra: oh that's my mistake. the .vs folder was previously in the gitignore, but it doesn't match against the new rule. i'll remove it now. thanks

cijothomas (approved):
perhaps a good time to delete nuget.exe. We just need one in the repo. (checked in binaries is bad practice anyways)

TimothyMothra:
Let's delete the binaries after the migration as part of the consolidated step.
When I start authoring new build definitions I'll confirm it's not in use.

